PR #14, opened 2020-10-26
added level as enum

gperonato:


db/add_dataset_db.sql
```diff
@@ -10,6 +10,8 @@ ALTER DATABASE dataset SET search_path = public, postgis;
 
 CREATE EXTENSION IF NOT EXISTS postgis ;
 
+CREATE TYPE levl AS ENUM ('0', '1', '2', '3', '4', '9');
```
gcmalloc: here you could have:

NUTS0 NUTS1 NUTS2 COUNTRY LAU or similar right ?

gperonato: Yes, I was just reusing the codes from the original datasets (transformed from int to str), but now with enum we can also give more understandable names

gcmalloc (approved):

PR #22, opened 2020-10-29
Use SIGINT as signal to `docker-compose stop db`

### LINKED ISSUE #20: README.md issue with db information
This instruction doesn't seem to work:

`the database on 127.0.0.1:5433`

I have a  `db_1        | LOG:  invalid length of startup packet` error. I cleaned everything and re-installed everything again. But the same issue arises

gcmalloc:
make sigint the default signal for stopping psql (see https://github.com/docker-library/postgres/issues/714), this makes the kill abort connection (idle ones included) and should make docker stop db more resilient.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -30,6 +30,7 @@ services:
     restart: always
     environment:
       POSTGRES_PASSWORD: example
+    stop_signal: SIGINT                 # Fast Shutdown mode
     volumes:
       - db-data:/var/lib/postgresql/data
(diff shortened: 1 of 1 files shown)


neodark:
What is the status of this PR ? Do you think we should keep this @gcmalloc ?

gcmalloc:
Totally ! I think it is a safer stop mechanism than the smart stop.

neodark (approved):

PR #40, opened 2020-11-14
Add flower

gcmalloc:
Flower allows us to monitor the celery task from outside the application and report metrics on them.


dianevg (approved):
I'm a bit out of my depth here :-).  But yes, monitoring the queue of the interactions by the cm seems a good idea!

PR #41, opened 2020-11-14
add a status code equal for easier test and more readable test output

gcmalloc:
Add a bit more test around the calculation module creation. 
Add an helper function to check the status code return of a flask test client response.


gcmalloc:
This is depending on # #38 for adding timeout in the celery inspect commands.

cedricserugendo (approved):

PR #90, opened 2020-11-26
Implement multiply test

cedricserugendo:


gcmalloc:
I fixed the following:

Projection of the geojson geometry according to the raster projection, this is done trough pyproj and shapely which I added as dependencies.
Only apply the factor in the resulting statistics and not on the entire raster.
Remove the usage of zonal_stat in the test and rewrite the test that doubles a raster and compare it to the original.

gcmalloc (approved):

PR #93, opened 2020-11-26
add readme for frontend and cm

gcmalloc:


cm/README.md
```diff
@@ -0,0 +1,24 @@
+base directory for the calculation modules (cm).
+
+A calculation module is an asynchronous slow task (from seconds to hours of
+latency for the answer). A calculation module take a set of raster, a selection and an optional set of parameters
```
dianevg: takes

cm/README.md
```diff
@@ -0,0 +1,24 @@
...
+Then you can connect to the redis queue. The multiply cm is given as an example.
+
+We strongly encourage you to write unittest that verify the good working of a calculation module separated from the enermaps2 api.
+
+# Create a new cm
+
+Each calculation module is a subdirectory in this directory.
+
+It must contain:
+* A setup.cfg symlink to the cm/setup.cfg file.
```
dianevg: A setup.cfg symlink to the cm/setup.cfg file - it might be need a bit more info (for me at least but I might just miss vocabulary)

dianevg (approved):

cm/README.md
gcmalloc: A very good point, it seems that on windows git symlink are pretty convoluted to use (rely on a configuration setting). 

I will put a reference to it for the moment, if someone else ask the question I will change it to copying and maintaining the copy up to date.

PR #95, opened 2020-11-26
change the approach to the frontend selector

gcmalloc:
Change the approach we use for updating the layers in leaflet. 
We still keep the map component in the App.svelte, but we now do selective update based on the layers that are currently present in the map and the one that are selected. We then sync the two states.

This PR also set the ZIndex for the overlay and selection layer by always putting the selection layer on top. As it doesn't make sense to have the selection layer below the rasters for example.


dianevg (approved):
Yes, it seems to work also! The selected layer should be the one on top.

PR #116, opened 2020-12-04
Migrate the getmap to be compatible with WMS 1.3.0

### LINKED ISSUE #57: Investigate the error from qgis
When QGIS is hitting the wms GetCapabilities endpoint, qgis return an error:

```
Invalid Layer: WMS provider Cannot calculate extent Raster layer Provider is not valid (provider: wms, URI: crs=ESPG:3857&dpiMode=7&format=image/png&layers=lau.zip&styles&url=http://127.0.0.1:7000/api/wms
```

It may be due to the dpiMode query parameter, or to the absence of style in the getcapability endpoint.

gcmalloc:


dianevg (approved):
ok, all good for me!

PR #128, opened 2020-12-11
Graph dataset

cedricserugendo:
adding scatter graph for visualisation of CM output


cm/multiply/multiply_raster.py
```diff
@@ -62,4 +73,22 @@ def rasterstats(geojson, raster_path, factor, stats=None):
 
 
 if __name__ == "__main__":
-    val_multiply = rasterstats("selection_shapefile.geojson", "GeoTIFF_test.tif", 2)
+    import json
+    import os
```
gcmalloc: import should be at the top of the file.

cm/multiply/multiply_raster.py
```diff
@@ -62,4 +73,22 @@ def rasterstats(geojson, raster_path, factor, stats=None):
...
+
+    CURRENT_FILE_DIR = os.path.dirname(os.path.abspath(__file__))
```
gcmalloc: this belong at the top of the file too.

frontend/Dockerfile
```diff
@@ -6,6 +6,7 @@ COPY src/package.json src/package-lock.json ./
 RUN npm install
 COPY src/ .
 RUN npm run build
+RUN npm install chart.js --save
```
gcmalloc, line 9: This should be in package.json and not here.

frontend/src/public/index.html
```diff
@@ -11,6 +11,8 @@
 	<link rel='stylesheet' href='/build/bundle.css'>
 	<link rel='stylesheet' href='/build/vendor.css'>
 
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/Chart.js/2.9.4/Chart.min.js" integrity="sha512-d9xgZrVZpmmQlfonhQUvTR7lMPtO7NkZMkA0ABN3PHCbKA5nqylQ/yWlFAyY6hYgdF1Qh6nYiuADWwKB4C2WSw==" crossorigin="anonymous"></script>
```
gcmalloc: if this is imported in Chart.svelte and installed with npm, it will be bundled in the bundle.js

PR #131, opened 2020-12-16
Cleanup cm output, create a module for shared function between CMs

gcmalloc:
Implement a python module to be used as base for the cms.

settle down on one format for the output of calculation modules and write a validator for it.

Implement a merge of feature when running the multiply plugin to have a summary accross all selection.
Implement a CDF calculation for the plot and use the previous definiton of the output of a calculation module to output a xy graph.


gcmalloc:
This PR needs to be broken down:

CM merging featureset into a single large feature.
base module for cm implementation.

cedricserugendo (approved):

PR #151, opened 2021-01-06
add delete button for CMTask

### LINKED ISSUE #140: Add a button to delete a particular CM output

gcmalloc:


knackbord:
Is that the "cancel task" button? because I do not see any other...

gcmalloc:
should be a small X at the top of the CMTask, you can reload the frontend
by running:

docker-compose up -d --build frontend

That should fix the problem.

knackbord:
Found the "X", I just did not see it before.
But we need to be able to close the CM as well. so it works but I would expect the X to bigger and in a different color or in square on the top right side not on the left

gcmalloc:
sounds good.
closing the cmpane is handled in https://github.com/enermaps/enermaps/issues/152. 
We may repass on this layout when we have more cm to display.

gcmalloc:
Restyled the button so that it is a clear that it is a close button. 
We can reuse the same style for toggling the CMList in another PR.

gcmalloc:
See 
2021-01-07-150748_346x625_scrot

gcmalloc:
see https://github.com/enermaps/enermaps/pull/154 for the button to close the CM panel

cedricserugendo (approved):

PR #172, opened 2021-02-04
put the try/expect into output_raster function

cedricserugendo:


neodark (approved):

neodark:
@cedricserugendo : to check why some tests fail

cm/base/BaseCM/cm_output.py
```diff
@@ -72,5 +74,9 @@ def validate(output: Dict) -> Dict:
...
     """Add a raster to the api"""
     files = {"file": (raster_name, raster_fd, "image/tiff")}
-    resp = requests.post(API_URL + "api/geofile/", files=files)
-    return resp.ok
+    try:
+        resp = requests.post(API_URL + "api/geofile/", files=files)
+        return resp.ok
+    except ConnectionError:
+        logging.error("Error during the post of the file.")
+        return False
```
melhk, line 82: The API should always return status codes, not True or False.

neodark: Yes this should be fixed as well. Thanks @melhk for pointing that out!

neodark: On a second thought, maybe we don't need status codes here as it's not an API.
This needs to be investigated more deeply:
* What does it mean when we return an error from the CM ?
* What is the procedure when everything works fine?
* What are the expected outputs for all cases ?

PR #174, opened 2021-02-11
modify output_raster to return status code

cedricserugendo:


melhk (commented):
We have to respond to Flavio's last comment before merging this code.

melhk (approved):

PR #176, opened 2021-02-23
Basic (manual) integration of three datasets and sample sql queries

gperonato:
See `README.md` in `/di `and `/db`


neodark:
This is just a basic remark. Please don't take it in another way as we are in the process of improving things, we can all learn about it:

My 2 cents here is the following:

- The size of this PR is huge... Maybe too much
- I'd suggest to have smaller PR to help a proper review
- Let's try and have more "atomic commits" in the sense that a commit should reflect a particular change
- The commits description should hold a clearer description ("small edits", "modifs", ...) otherwise it's not possible to understand the changes

Could we all improve on this part please as multiple developers will work on this?
Thanks to all!

gperonato:
I am sorry for the huge PR. This PR is the initial commit of the data integration process. Future PRs will be smaller.
As agreed in the last meeting, I'll write here some details about the review to be conducted.
The goal is to review the main features and not the code in detail.
There are three python scripts that need to be manually executed in the Docker app to integrate three datasets into the db. Here are the steps:
- Make sure that the db is running
```
docker-compose up -d db
```
- Make sure that the db is empty; this query should give a 0 count:
```
enermaps % docker exec -it enermaps_db_1 psql -h 127.0.0.1 -p 5432 -U test dataset -c "select count(ds_id) from datasets;"
```
- Activate the `data-integration` service
```
docker-compose up -d data-integration
```
- Insert the first dataset, by running this command:
```
docker-compose exec data-integration python3 admin_units.py
```

- After the integration is completed, test this query:
```
docker exec -it enermaps_db_1 psql -h 127.0.0.1 -p 5432 -U test dataset -c 'select "FID" from spatial where ds_id = 0 LIMIT 2;'
```
It should return:
```
 FID 

 AE
 AD
(2 rows)
```
- Insert the second dataset, by running this command:
```
docker-compose exec data-integration python3 getHotMaps_raster.py --select_ds_ids 31
```

- Execute this command, which should list the tiff files that have been inserted:
```
docker-compose exec data-integration ls data/31
```
- Excecute this query, which should give the name of the first two datasets:
```
docker exec -it enermaps_db_1 psql -h 127.0.0.1 -p 5432 -U test dataset -c 'select "FID" from data where ds_id = 31 LIMIT 2;
```
- Insert the third dataset by running
```
docker-compose exec data-integration python3 getJRC_GEOPP_DB_csv.py
```
- Run this query:
```
docker exec -it enermaps_db_1 psql -h 127.0.0.1 -p 5432 -U test dataset -c 'select "variable", "value" from data where ds_id = 2 LIMIT 2;' 
```

    which should return:

```
    variable    | value 

 gross_cap_ele |    60
 ini_cap_ele   |    60
```
- I would appreciate if you could check the documentation (`README.MD` files in the `di` and `db` directories) to check whether they are consistent and clear enough.

db/README.md
```diff
@@ -1,10 +1,42 @@
-database for the enermaps2 project.
+# Database for the enermaps project
 
 We use postgresql with postgis. The initial database schema 
 can be found in db/add_dataset_db.sql
 
+The sql schema will be loaded only if the `db-data` volume is not already present.
+You can use `docker-compose down --volumes` and `docker volume rm db-data` before running the app to load the schema and start from an empty db.
```
gcmalloc, line 7: This information is already in the README at the root of the directory, I see two solution there:
* Either you have a documentation specific to the db that drop only the db volume
* Or you rely on the README at the root of the repository and add a reference to it

db/README.md
```diff
@@ -1,10 +1,42 @@
...
 ```
-docker exec -it enermaps2_db_1 psql -h 127.0.0.1 -p 5432 -U dataset dataset
+docker exec -it enermaps_db_1 psql -h 127.0.0.1 -p 5432 -U dataset dataset
+```
+
+## Sample queries
+
+- Vector dataset, here with `ds_id = 2`
+
+```
```
gcmalloc: Add sql after the three backtick so you get syntax highlighting

```sql 
select * from base limit 1000
```

db/README.md
```diff
@@ -1,10 +1,42 @@
...
+		json_object_agg(variable, value),
+		fields,
+		time, dt, z, data.ds_id, metadata, geometry
+       FROM data
+INNER JOIN spatial ON data."FID" = spatial."FID"
+INNER JOIN datasets ON data.ds_id = datasets.ds_id
+WHERE data.ds_id = 2
+GROUP BY data."FID", time, dt, z, data.ds_id, fields, geometry, metadata
+ORDER BY data."FID";
 ```
```
gcmalloc, line 30: Newline after the backticks

db/add_dataset_db.sql
```diff
@@ -25,19 +25,36 @@ CREATE TABLE public.spatial
...
+    REFERENCES datasets(ds_id)
+    ON DELETE CASCADE
+;
+
+ALTER TABLE data
+    ADD CONSTRAINT fk_ds_id
+    FOREIGN KEY(ds_id) 
+    REFERENCES datasets(ds_id)
+    ON DELETE CASCADE
+;
```
gcmalloc, line 60: Note that you could do some database initialisation here too, for example I wouldn't expect the database to be initialized without statistical units, maybe including this in the db initialization would make sense ?

di/Dockerfile
```diff
@@ -0,0 +1,38 @@
+FROM ubuntu:20.04
+# SET TZ for GDAL
+ENV TZ=Europe/Zurich
+RUN ln -snf /usr/share/zoneinfo/$TZ /etc/localtime && echo $TZ > /etc/timezone
+RUN apt-get update && \
+    apt-get --yes install python3 python3-pip &&\
+    rm -rf /var/cache/apt/archives/
+
+
+RUN apt-get --yes install postgresql-client
```
gcmalloc: those should be bundled with the install command on line 6.
If you want to keep the command separated, don't forget to remove the cache with

```sh
rm -rf /var/cache/apt/archives/
```

di/Dockerfile
```diff
@@ -0,0 +1,38 @@
...
+
+RUN apt-get update && apt-get --yes install gdal-bin && apt-get --yes install libgdal-dev
```
gcmalloc: Don't need the update there, when you run the update it will locally fetch a database of packages. You could use the command on line 6 and bundle the install together.

di/Dockerfile
```diff
@@ -0,0 +1,38 @@
...
+RUN export CPLUS_INCLUDE_PATH=/usr/include/gdal
+RUN export C_INCLUDE_PATH=/usr/include/gdal
```
gcmalloc: Use the ENV docker statement (see https://docs.docker.com/engine/reference/builder/#env)

```dockerfile
ENV C_INCLUDE_PATH /usr/include/gdal
```

di/Dockerfile
```diff
@@ -0,0 +1,38 @@
...
+
+# RUN apt-get --yes install git
```
gcmalloc: To be removed.

di/Dockerfile
```diff
@@ -0,0 +1,38 @@
...
+# RUN cd datapackage-py && pip3 install -e .
+
+
+# RUN git clone --branch raster https://github.com/enermaps/dataflows.git
+# RUN cd dataflows && pip3 install -e .
+
+
+# CMD ["python3", "admin_units.py"]
+
+# CMD ["python3", "getNEWA.py"]
```
gcmalloc: I think the commented approach looks good, by default running the container will just import datasets. Why did you remove it ?

di/README.md
```diff
@@ -0,0 +1,19 @@
...
+    `docker-compose exec data-integration python3 admin_units.py`
+
+  - 25: NEWA data
+  	`docker-compose exec data-integration python3 getNEWA.py`
+
+  - 2: JRC Geothermal power plants
+  `docker-compose exec data-integration python3 getJRC_GEOPP_DB_csv.py`
+  	
+
+Note: you need to start both the db service (`docker-compose up -d db`) and the data-integration service (`docker-compose up -d data-integration`) before running the pipelines.
```
gcmalloc: It's a nice readme !

di/admin_units.py
```diff
@@ -0,0 +1,86 @@
...
+}
+
+
+def get(datasets=datasets, crs="EPSG:4326"):
+    """
+    Retrieve NUTS, LAU and countries from GISCO API and make a single, consistent GDF.
+
+    Parameters
+    ----------
+    datasets : dict
```
gcmalloc: Those are good example of type annotation, you can specify a type directly in your method, for example:
```
def get(datasets: dict=datasets, crs: str="EPSG:4326") -> gpd.GeoDataFrame:
```

di/admin_units.py
```diff
@@ -0,0 +1,86 @@
...
+    crs : TYPE, optional
+        Spatial Reference System. The default is "EPSG:4326".
+
+    Returns
+    -------
+    admin_units : GeoDataFrame
+        Table with all administrative units.
+
+    """
+    print("Downloading countries...")
```
gcmalloc: nit. You can use logging instead.

db/README.md
gperonato: No, it's not, at least the `docker volume rm db-data` part: it took me a while to understand that rebuilding the db was not enough.  Anyway, I don't see a major problem in repeating some concepts. But indeed we could avoid having a separate readme for each service and group all doc in the main readme.

di/admin_units.py
```diff
@@ -0,0 +1,86 @@
...
+    countries = gpd.read_file(datasets["countries"].format(crs.split(":")[-1]), crs=crs)
```
gcmalloc: I feel like the crs parsing could be done by a lib, or at least the split could be separated high up in the function.

di/admin_units.py
```diff
@@ -0,0 +1,86 @@
...
+    admin_units.LEVL_CODE = admin_units.LEVL_CODE.replace({0:"country",1:"NUTS1",2:"NUTS2",3:"NUTS3",4:"LAU"})
+
+    admin_units.crs = crs
+    return admin_units
+
+
+
+
+if __name__ == "__main__":
+    host = "enermaps_db_1"
```
gcmalloc: inside the docker compose you can directly talk to `db`, it would be nice to have a set of common parameters between all import script.

Alernatively you could use the environment variable specified in the .env at the root of this repository.

di/admin_units.py
```diff
@@ -0,0 +1,86 @@
...
+    port = 5432
+    ds_id = 0
+    admin_units = get(datasets, crs="EPSG:3035")
```
gcmalloc: Could you move that const at the top of the file ?

di/datasets.csv
```diff
@@ -0,0 +1,57 @@
+ID,Group (datasets of the same group are similar*),Project,Title (with Hyperlink),Description (in brief),Metadata URLs,Methodology (URL to methodology descriptions),Methodology (brief description),Accuracy,Completeness,Level,Spatial Granularity,Identifier,Identifier Type,Creator,Object,Publisher,Publication Date,Publication Year,Temporal Granularity,Time references (time data refers),URLs,Content (keywords),Origin,Geographical extension,Projection system,Access conditions,License,Terms of use,Availability,Resource type,Data format,"Size of file (raw, compressed in parentheses)",Other relevant information
```
gcmalloc: latin1 is not really a recommended encoding for files, could you translate this file to use utf-8 instead ?

di/getJRC_GEOPP_DB_csv.py
```diff
@@ -0,0 +1,137 @@
+from dataflows import Flow, load, dump_to_path, dump_to_zip, printer, add_metadata, update_package, update_resource, update_schema, set_primary_key
+from dataflows import sort_rows, filter_rows, find_replace, delete_fields, select_fields
+from dataflows import add_computed_field, unpivot, set_type, validate, unpivot, duplicate, dump_to_sql
+from datapackage import Package
+import json
```
gcmalloc: Run isort on those.

di/getJRC_GEOPP_DB_csv.py
```diff
@@ -0,0 +1,137 @@
...
+
+        # Resource 0
+        unpivot(unpivoting_fields, extra_keys, extra_value, regex=True,  resources=0),
+        delete_fields(['geometry'],resources=0),
+        find_replace([{"name": "value", "patterns":[{"find":"NULL","replace":""}]}], resources=0),
+        # update_schema(None, missingValues=["NULL", ""], resources=0),
+        set_type('value',type="number", resources=0),
+        # validate()
+        # printer(num_rows=10),
+    )
```
gcmalloc: that's a super long method definition that got me lost around line 66, any way that we could simplify this ?

di/getJRC_GEOPP_DB_csv.py
```diff
@@ -0,0 +1,137 @@
...
+        # dump_to_sql({"spatial":{"resource-name":"jrc_geopp_db_spatial", "mode":"append"},
+        #             "data":{"resource-name":"jrc_geopp_db", "mode":"append"}},
+        #             engine="postgresql://test:example@{host}:{port}/dataset".format(host=host,port=port))
+        ).process()
+    
+    import pandas as pd
+    import geopandas as gpd
+    import utilities
+    from shapely import wkt
+    import shutil
```
gcmalloc: I don't understand the reasoning behind putting those import here and not at the top of the file.

di/getJRC_GEOPP_DB_csv.py
```diff
@@ -0,0 +1,137 @@
...
+    spatial['geometry'] = spatial['geometry'].apply(wkt.loads)
+    spatial = gpd.GeoDataFrame(spatial)
+    spatial.crs = "EPSG:4326"
+    spatial = spatial.to_crs("EPSG:3035")
+    spatial["ds_id"] =  ds_id
+    utilities.toPostGIS(spatial,"postgresql://test:example@{host}:{port}/dataset".format(host=host,port=port), schema="spatial")
+    
+    shutil.rmtree("data/jrc_geopp_db/")
+    
+
```
gcmalloc: this file need a pass of black.

di/getNEWA.py
```diff
@@ -0,0 +1,115 @@
...
+
+
+import pandas as pd
+import requests
+import subprocess
+from osgeo import gdal, osr
+import os
+import utilities
+
+def get(time="2009-12-30", timestep=0.5, heights=[100], variable="PD", Run=False):
```
gcmalloc: shouldn't time be a datetime.date instead ?

di/getNEWA.py
```diff
@@ -0,0 +1,115 @@
...
+            )
+    else:
+        timeseries = [time]
+
+    for time in timeseries:
+        time = pd.to_datetime(time)
+        year = time.year
+        month = time.month
+        day = time.day
+        height_levels = [50, 75, 100, 150, 200, 250, 500]
```
gcmalloc: this belong at the top of the file in a const.

```sh
HEIGHT_LEVELS = [50, 75, 100, 150, 200, 250, 500]
```

di/getNEWA.py
```diff
@@ -0,0 +1,115 @@
...
+            filename = "NEWA-{year}-{month}-{day}_{height}m_{ts}min.nc".format(
+                year=year, month=month, day=day, height=height, ts=int(timestep * 60)
+            )
+            # files.append(filename)
+            if Run:
+                print("Requesting", URL)
+                # Issue about the server pretending to send gzip, patched as here:
+                # https://github.com/psf/requests/issues/3849#issuecomment-277196788
+                r = requests.get(URL, allow_redirects=True, headers={'Accept-Encoding': 'identity'})
+                open(filename, "wb").write(r.content)
```
gcmalloc: use with here

```python
with open(filename, 'wb') as f:
    f.write(r.content)
```

docker-compose.yml
```diff
@@ -60,6 +60,12 @@ services:
     env_file:
       .env
 
+  data-integration:
+    build:
+      context: ./di
```
gcmalloc: I fthink you could rename the directory data-integration to match the name of the docker image.

gcmalloc (commented):
Just a couple non-blocking comments, the python files need some linting.

db/README.md
gcmalloc: even in this section: https://github.com/enermaps/enermaps#cleanup ? `docker-compose down -v` seems to be pretty similar to the command you proposed, on my machine it prints the following:

```sh
$ docker-compose down -v                                                                                 
Removing network enermaps2_default
Removing volume enermaps2_upload-data
Removing volume enermaps2_db-data
Removing volume enermaps2_data
```

Is there something I missed ?

gperonato: Ah I see your point. For some reason, in my machine it doesn't remove the `db-data` volume...
```
$  docker-compose down -v
Stopping enermaps_api_1              ... done
Stopping enermaps_frontend_1         ... done
Stopping enermaps_data-integration_1 ... done
Stopping enermaps_db_1               ... done
WARNING: Found orphan containers (enermaps_cm-countclass_1, enermaps_thumbnails_1) for this project. If you removed or renamed this service in your compose file, you can run this command with the --remove-orphans flag to clean it up.
Removing enermaps_api_1              ... done
Removing enermaps_cm-multiply_1      ... done
Removing enermaps_flower_1           ... done
Removing enermaps_redis_1            ... done
Removing enermaps_cm-geofile_1       ... done
Removing enermaps_frontend_1         ... done
Removing enermaps_data-integration_1 ... done
Removing enermaps_db_1               ... done
Removing network enermaps_default
Removing volume enermaps_upload-data
ERROR: remove enermaps_upload-data: volume is in use - [9c24dbd20c779b98360e1e8da7d4b66fd978821f6a4594419d4795a7d75f6904]
```

gperonato:
Thanks a lot @gcmalloc for the very detailed review and useful comments! I'll go through each of them in the next days!

gcmalloc:
No problem, I will see if I can get some time for reviewing the pipeline steps themself.

Don't be afraid of doing smaller cl, they make the reviewer job faster and easier.

db/README.md
gcmalloc: Sounds like another problem, see:
```
WARNING: Found orphan containers (enermaps_cm-countclass_1, enermaps_thumbnails_1) for this project. If you removed or renamed this service in your compose file, you can run this command with the --remove-orphans flag to clean it up.
```
it means that you have countainers that are still running in this project that are not currently in the docker-compose file. You can check for those and kill them using:

```sh
docker ps | grep -v CONTAINER | awk '{ print $1  }' | xargs docker stop
```

di/Dockerfile
gperonato: Yes, it's actual a reminder of how the pipeline should look like (will add a comment about this). For testing, I prefer to keep a manual integration.

di/admin_units.py
gperonato: Nice!

di/getJRC_GEOPP_DB_csv.py
gperonato: I will rewrite it using (Geo)Pandas only, should simplify things

db/add_dataset_db.sql
gperonato: For now I prefer to keep only the schema in db, and all data integration is done as a separate service.

cm/countclass/countclass.py
```diff
@@ -0,0 +1,58 @@
...
+    stat_done = time()
+    ret = dict()
+    ret["graphs"] = []
+    ret["geofiles"] = []
+    ret["values"] = stat
+
+    logging.info("We took {!s} to calculate the area".format(stat_done - start))
+
+    validate(ret)
+    return ret
```
gcmalloc: Could you move the new cm into another pull request ? that would make that pull request way smaller and easier to digest.

cm/countclass/countclass.py
```diff
@@ -0,0 +1,58 @@
...
+    for feature in geojson["features"]:
+        geometry = feature["geometry"]
+        geoshape = shape(geometry)
+        projected_shape = transform(project, geoshape)
+        geometries.append(projected_shape)
+    merged_geometries = cascaded_union(geometries)
+
+    with rasterio.open(raster_path) as src:
+        raster, raster_transform = rasterio.mask.mask(src, [merged_geometries], crop=True)
+        print(src.meta)
```
gcmalloc: You could use the logging module here instead see https://docs.python.org/3/howto/logging.html

data-integration/Dockerfile
```diff
@@ -0,0 +1,27 @@
+FROM ubuntu:20.04
+# SET TZ for GDAL
+ENV TZ=Europe/Zurich
+RUN ln -snf /usr/share/zoneinfo/$TZ /etc/localtime && echo $TZ > /etc/timezone
+
+RUN apt-get update && \
+    apt-get --yes install python3 python3-pip &&\
+    apt-get --yes install postgresql-client &&\
+    apt-get --yes install gdal-bin &&\
+    apt-get --yes install libgdal-dev &&\
```
gcmalloc: you could do 


```sh
RUN apt-get --yes install postgresql-client \
    gdal-bin \
    libgdal-dev &&\
    rm -rf /var/cache/apt/archives/
```
which would be a bit faster

data-integration/admin_units.py
```diff
@@ -0,0 +1,98 @@
...
+    crs : str, optional
+        Spatial Reference System. The default is "EPSG:4326".
+
+    Returns
+    -------
+    admin_units : GeoDataFrame
+        Table with all administrative units.
+
+    """
+    source_crs_code = crs.split(":")[-1]
```
gcmalloc: what do you think about using pyproj here instead ?

data-integration/admin_units.py
```diff
@@ -0,0 +1,98 @@
...
+
+    # Create consistent columns across ds
+    lau = lau.rename({"LAU_NAME": "NAME"}, axis=1)
+    lau["LEVL_CODE"] = 4
+
+    nuts = nuts.rename({"NAME_LATN": "NAME"}, axis=1)
+
+    countries = countries.rename({"CNTR_NAME": "NAME"}, axis=1)
+    countries = countries.rename({"CNTR_ID": "CNTR_CODE"}, axis=1)
+    countries["LEVL_CODE"] = 0
```
gcmalloc, line 72: something that can be done here is split this function into three separate call:

0) nuts download and formatting
0) lau download and formatting
0) country download and formatting

docker-compose.yml
```diff
@@ -60,6 +60,12 @@ services:
     env_file:
       .env
 
+  data-integration:
+    build:
+      context: ./data-integration
+    volumes:
+      - data:/data
```
gcmalloc, line 67: do you also want the env here to have the database configuration ?

data-integration/utilities.py
```diff
@@ -0,0 +1,218 @@
...
+    ds_id : int.
+    dbURL : str, optional
+        The default is "postgresql://test:example@localhost:5433/dataset".
+    Returns
+    -------
+    datatpackage: json string.
+    """
+    engine = sqla.create_engine(dbURL)
+    df = pd.read_sql(
+        "SELECT * FROM datasets WHERE ds_id = {};".format(ds_id), con=engine
```
gcmalloc: Carefull here, formatting sql string that way is usually a bad idea as it leads to sqlinjection quite easily. Meaning you can have ds_id being an sql command and the command will be happilly executed. You may want to change the call to 

```python
        pd.read_sql("SELECT * FROM datasets WHERE ds_id = ?;", params=(ds_id,), con=engine)
```

data-integration/utilities.py
```diff
@@ -0,0 +1,218 @@
...
+    )
+    if len(df) > 0:
```
gcmalloc: In python you can directly call if on df

```python
if df:
...
```

data-integration/utilities.py
```diff
@@ -0,0 +1,218 @@
...
+    """
+    Retrieve the datapackage.
+    Parameters
+    ----------
+    ds_id : int.
+    dbURL : str, optional
+        The default is "postgresql://test:example@localhost:5433/dataset".
+    Returns
+    -------
+    datatpackage: json string.
```
gcmalloc: Maybe replace by "Optional datapackage" or "datapackage or None" as we can also return None

db/Dockerfile
```diff
@@ -1,5 +1,5 @@
 FROM camptocamp/postgres:9.6
-
+# RUN rm -r /var/lib/postgresql/data/*
 ENV POSTGRES_USER test
 ENV POSTGRES_DB test
-ADD *.sql /docker-entrypoint-initdb.d/
+ADD add_dataset_db.sql /docker-entrypoint-initdb.d/add_dataset_db.sql
```
gcmalloc: maybe move the initial script into a subdirectory instead such as db/init ?

db/Dockerfile
```diff
@@ -1,5 +1,5 @@
 FROM camptocamp/postgres:9.6
-
+# RUN rm -r /var/lib/postgresql/data/*
```
gcmalloc: This can be removed, the data from the base image (camptocamp/postgres) will always be empty.

db/README.md
gcmalloc: The command is still there, do you want to remove it as it cannot be made specific to this container ?

gcmalloc (commented):
A couple more comments, files still need a pass of isort.

cm/countclass/countclass.py
gperonato: Yes, sorry, it was supposed to (it was added here by mistake)

data-integration/admin_units.py
gperonato: Yes, but it think it is easier like that, they populate the same table anyway

data-integration/utilities.py
gperonato: Thanks :-)

data-integration/utilities.py
gperonato: I don't think so with a DataFrame. It gives ```ValueError: The truth value of a DataFrame is ambiguous.``

db/Dockerfile
gperonato: Not sure I understand...

gperonato:
I have addressed most of Malik's comments (left open the others).
@melhk @neodark Can you please check [this comment ](https://github.com/enermaps/enermaps/pull/176#issuecomment-801755807) to complete the review?

melhk:
> I have addressed most of Malik's comments (left open the others).
> @melhk @neodark Can you please check [this comment ](https://github.com/enermaps/enermaps/pull/176#issuecomment-801755807) to complete the review?

Hello @gperonato, I'm on vacation at the moment and can check it on Wednesday next week. 

Yesterday I had an error (beautifulsoup library missing) when running the command `docker-compose exec data-integration python3 admin_units.py`

But I see that you have just modified the requirements.txt file, so I think it is already solved. I haven't gone further yet.

data-integration/utilities.py
gcmalloc: Oh didn't notice it was a dataframe, what about `not df.empty` then ?

docker-compose.yml
gperonato: I have now set the db parameters in the `.env` file, I don't know whether this is what you meant

data-integration/utilities.py
```diff
@@ -165,18 +163,18 @@ def datasetExists(
...
     bool
     """
+    tables = ["datasets", "spatial", "data"]
     engine = sqla.create_engine(dbURL)
     lengths = []
     for table in tables:
         with engine.connect() as con:
             rs = con.execute(
-                "SELECT COUNT(*) FROM {} WHERE ds_id = {}".format(table, ds_id)
+                "SELECT COUNT(*) FROM {} WHERE ds_id = %(ds_id)s;".format(table),
```
gperonato: @gcmalloc any suggestion on how to safely (and simply) pass the table name to the query with sqlalchemy? The Pythonic interface of sqlalchemy seems an overkill here. Anyway, I provisionally set the table list inside the function, so this should also be safe enough, right?

gcmalloc: You could use 

engine.has_table(table) as a check, you could also use engine.table_names() to get the list of tables. You could also call the underlying escape of psql (sql.Identifier see https://stackoverflow.com/questions/13793399/passing-table-name-as-a-parameter-in-psycopg2). Another solution would be to rewrite the function to directly include the table and make either an union of all matches or add the three count from the separate tables.

Alternatively, you can use sqlalchemy function directly, but not using the ORM feature is mostly getting in your way there. Maybe you want to use https://www.psycopg.org/ directly instead ?

melhk:
>     * d clear

Hello @gperonato, I'm trying to follow your procedure, but I'm stuck at the command `docker-compose exec data-integration python3 admin_units.py`.

Your last changes solved the import error, but now I get: 
```
  File "admin_units.py", line 111, in <module>
    utilities.toPostGIS(
  File "/di/utilities.py", line 152, in toPostGIS
    gdf.to_postgis(schema, db_engine, if_exists="append", index=False)
  File "/usr/local/lib/python3.8/dist-packages/geopandas/geodataframe.py", line 1097, in to_postgis
    geopandas.io.sql._write_postgis(
  File "/usr/local/lib/python3.8/dist-packages/geopandas/io/sql.py", line 388, in _write_postgis
    raise ValueError(msg)
ValueError: The CRS of the target table (EPSG:4326) differs from the CRS of current GeoDataFrame (EPSG:3035).
ERROR: 1
```

data-integration/utilities.py
melhk: Are you sure this function does what you want?
In the description you say "..exist in all tables" but you compare the results of the queries with an "or" operator!

gperonato:
Hi @melhk, did you remove the database volume before the data-integration? I think you have an old version of the db where the EPSG was 4326. I think
```
docker-compose down -v
docker-compose up --build -d db
```
should work

data-integration/utilities.py
```diff
@@ -165,18 +163,18 @@ def datasetExists(
...
+                {"ds_id": ds_id},
             )
             for row in rs:
```
melhk: Using a for loop is useless in this case since the query gives only one line result, and will lead to bugs if the query is changed by one giving several line results.

gperonato: The query gives more than one result, that's why a for loop is used. But I agree that is not elegant at the moment. I am open to suggestions in case. Anyway, I am trying to rebuild this function using psycopg2 (see thread with Malik), so this might be disappear as well.

gperonato: You are right, the description is not clear. I did want an `or` operator here.

data-integration/utilities.py
```diff
@@ -165,18 +163,18 @@ def datasetExists(
...
                 count = row[0]
```
melhk: Then the count variable is reassigned on each iteration and only the last value is used ... is it what you want??

gperonato: Yes, (sorry my previous answer was not clear): for some reason the result is an iterable with only one item

melhk:
The command `docker-compose exec data-integration python3 admin_units.py` works now that I did `docker-compose down -v`, you should add this to your instructions.

But now I'm stuck at the command `docker exec -it enermaps_db_1 psql -h 127.0.0.1 -p 5432 -U test dataset -c 'select "FID" from spatial where ds_id = 0 LIMIT 2;'`

Whit the error : 
```
ERROR:  column "fid" does not exist
LINE 1: select FID from spatial where ds_id = 0 LIMIT 2;
```

cm/multiply/multiply_raster.py
```diff
@@ -9,6 +9,13 @@
 from rasterstats import zonal_stats
 from shapely.geometry import shape
 from shapely.ops import cascaded_union, transform
+# print(os.environ)
+# print("")
+# print("")
+# print(os.listdir("/usr/local/lib/python3.8/dist-packages/pyproj"))
+#os.environ['PROJ_LIB'] = '/usr/local/share/proj'
+# print(os.environ)
+# print(os.environ)
```
melhk, line 19: Please, don't commit useless commented lines...

gperonato:
> The command `docker-compose exec data-integration python3 admin_units.py` works now that I did `docker-compose down -v`, you should add this to your instructions.

Sorry, it was in the readme, but I forgot to repeat that in the instructions.

> But now I'm stuck at the command `docker exec -it enermaps_db_1 psql -h 127.0.0.1 -p 5432 -U test dataset -c 'select "FID" from spatial where ds_id = 0 LIMIT 2;'`
> 
> Whit the error :
> 
> ```
> ERROR:  column "fid" does not exist
> LINE 1: select FID from spatial where ds_id = 0 LIMIT 2;
> ```
It is weird, the query works here. Can you check which columns you have in your table? 
`docker exec -it enermaps_db_1 psql -h 127.0.0.1 -p 5432 -U test dataset -c 'select * from spatial where ds_id = 0 LIMIT 1;'`

melhk:
> 
> 
> > The command `docker-compose exec data-integration python3 admin_units.py` works now that I did `docker-compose down -v`, you should add this to your instructions.
> 
> Sorry, it was in the readme, but I forgot to repeat that in the instructions.
> 
> > But now I'm stuck at the command `docker exec -it enermaps_db_1 psql -h 127.0.0.1 -p 5432 -U test dataset -c 'select "FID" from spatial where ds_id = 0 LIMIT 2;'`
> > Whit the error :
> > ```
> > ERROR:  column "fid" does not exist
> > LINE 1: select FID from spatial where ds_id = 0 LIMIT 2;
> > ```
> 
> It is weird, the query works here. Can you check which columns you have in your table?
> `docker exec -it enermaps_db_1 psql -h 127.0.0.1 -p 5432 -U test dataset -c 'select * from spatial where ds_id = 0 LIMIT 1;'`

This is the result:
` FID |           NAME           |      NAME_ENGL       | CNTR_CODE | LEVL_CODE | ds_id |`

and the table seems empty

cm/multiply/multiply_raster.py
gperonato: Ok, sorry for that. By the way, I had removed the CM from this PR. It shouldn't be there.

gperonato:
> This is the result:
> ` FID | NAME | NAME_ENGL | CNTR_CODE | LEVL_CODE | ds_id |`

It seems correct. And the FID is a table column, so I don't see why the previous query doesn't work. 

> and the table seems empty

Did you get any error on the data integration?

melhk:
> 
> 
> > This is the result:
> > ` FID | NAME | NAME_ENGL | CNTR_CODE | LEVL_CODE | ds_id |`
> 
> It seems correct. And the FID is a table column, so I don't see why the previous query doesn't work.
> 
> > and the table seems empty
> 
> Did you get any error on the data integration?

No, this was the output of the data integration:

```
INFO:root:Downloading countries...
INFO:root:Downloading NUTS...
INFO:root:Downloading LAU...
INFO:root:Done.
INFO:root:Loading to PostgreSQL...
INFO:root:Done.
INFO:root:Loading to PostGIS...
INFO:root:Done.
```

gperonato:
> > Did you get any error on the data integration?
> 
> No, this was the output of the data integration:
> 
> ```
> INFO:root:Downloading countries...
> INFO:root:Downloading NUTS...
> INFO:root:Downloading LAU...
> INFO:root:Done.
> INFO:root:Loading to PostgreSQL...
> INFO:root:Done.
> INFO:root:Loading to PostGIS...
> INFO:root:Done.
> ```

Can you investigate on that? I tried to rebuild the db, but cannot reproduce the problem.

data-integration/utilities.py
gperonato: I cleaned this a bit in the new commit (which uses psycopg2 directly).

data-integration/utilities.py
gperonato: Thank you @gcmalloc ! I ended up using psycopg2 directly, as it was already installed. I think it is much cleaner (and safer) now.
@melhk Thanks, I updated the description as well

melhk:
> 
> 
> > > Did you get any error on the data integration?
> > 
> > 
> > No, this was the output of the data integration:
> > ```
> > INFO:root:Downloading countries...
> > INFO:root:Downloading NUTS...
> > INFO:root:Downloading LAU...
> > INFO:root:Done.
> > INFO:root:Loading to PostgreSQL...
> > INFO:root:Done.
> > INFO:root:Loading to PostGIS...
> > INFO:root:Done.
> > ```
> 
> Can you investigate on that? I tried to rebuild the db, but cannot reproduce the problem.

With the help of Olivier, we found that putting quotes in your variable names can lead to some errors due possible differences of interpretation made by different shells...
Avoiding that (and capital letters) would be a good thing.

data-integration/getHotMaps_raster.py
```diff
@@ -0,0 +1,241 @@
...
+    for r in range(len(new_dp["resources"])):
+        if "temporal" in new_dp["resources"][r]:
+            time = pd.to_datetime(new_dp["resources"][r]["temporal"]["start"])
+        else:
+            time = None
+
+        if "temporal" in new_dp["resources"][r]:
+            time = pd.to_datetime(new_dp["resources"][r]["temporal"]["start"])
+        else:
+            time = None
```
melhk: Why are these lines doubled?

data-integration/getHotMaps_raster.py
```diff
@@ -0,0 +1,241 @@
...
+                "value": os.path.basename(new_dp["resources"][r]["path"]),
+                "time": time,
+                "z": None,
+                "unit": unit,
+                "dt": 8760,
+            }
+            rasters.append(raster)
+            # check statistics for each resource
+            if dp != None and "stats" in new_dp["resources"][r]:
+                if [dp["resources"][r]["stats"] != new_dp["resources"][r]["stats"]]:
```
melhk: This test has no effect, you are putting a "True" or a "False" in a list and then test if there is a list! 
The result is always true!

data-integration/getJRC_GEOPP_DB_csv.py
```diff
@@ -0,0 +1,256 @@
...
+    # Logic for update
+    if dp != None:
+        # check stats
+        ChangedStats = dp["resources"][0]["stats"] != new_dp["resources"][0]["stats"]
+        ChangedDate = dp["datePublished"] != new_dp["datePublished"]
+
+        if ChangedStats or ChangedDate:
+            logging.info("Data has changed")
+        elif force:
+            logging.info("Forced update")
```
melhk, line 95: Where is the forced update done?

data-integration/getJRC_GEOPP_DB_csv.py
```diff
@@ -0,0 +1,256 @@
...
+            ]
+        )
+        enermaps_data["Raster"] = False
+        enermaps_data["value"] = data["value"]
+        enermaps_data["unit"] = "MW"
+        enermaps_data["variable"] = data["variable"]
+        enermaps_data["fields"] = data["fields"]
+        # Constants
+        enermaps_data["unit"] = "MW"
+        enermaps_data["Raster"] = False
```
melhk, line 166: Why is this line the same as the line 155?

data-integration/utilities.py
```diff
@@ -0,0 +1,269 @@
...
+    chunk_size : integer, optional
+        The default is 128.
+    Returns
+    -------
+    None.
+    """
+    r = requests.get(url, allow_redirects=True, stream=True, timeout=timeout)
+    if len(append_path) > 0:
+        url = r.url + append_path
+        r = requests.get(url, stream=True, timeout=timeout)
```
melhk: I don't understand what you are doing here. Why are you doing two times the get request, once with an "append_path", what is this "append_path"?

You should also check that the string that you give to the get function is a valid url, and the http response of the get function otherwise you will have an error or write an empty file.

data-integration/utilities.py
gcmalloc: It looks better :)
count(*) will return a single line, you can use fetchone() to get the result. What do you think of the following:

```python
cursor = con.cursor()
cursor = cursor.execute("SELECT COUNT(*) FROM table WHERE ds_id = %(ds_id)s;", {'ds': ds})
result = cursor.fetchone()
if result:
  match_count = result[0]
```

gcmalloc:
Something you could potentially do is add the database load into the integration tests ran in https://github.com/enermaps/enermaps/blob/master/.github/workflows/pr_reviewed.yml. This would verify that each time a modification is done either to the loading script or to the database schema, the dataset loading is still working.

data-integration/getHotMaps_raster.py
gperonato: I don't understand your point. ChangedStats is not a a list. The logic is to check whether there is at least a resource with  some changes.

data-integration/getJRC_GEOPP_DB_csv.py
gperonato: It is done by existing continuing executing the code. On the contrary, if no forced update is called, the function will return three `None`s

melhk: So, if I understand correctly:
- If the data has changed (ChangedStats and/or ChangedDate set to True), you tell that the data has changed, then you make the forced update regardless if force is True or False.
- If the data has not changed and force is False, you tell that the data has not changed and return None
- If the data has not changed and force is True, you tell "forced update" and make the forced update

Do you really want the first case?

gperonato: Yes. The forced update will be mostly use while testing. In production, if the source dataset has changed, then it will be loaded into the db.

gperonato:
> Something you could potentially do is add the database load into the integration tests ran in https://github.com/enermaps/enermaps/blob/master/.github/workflows/pr_reviewed.yml. This would verify that each time a modification is done either to the loading script or to the database schema, the dataset loading is still working.

Yes, you are right. We should discuss about that at next sprint, it think it should be in another PR

data-integration/getHotMaps_raster.py
melhk: I am not talking about ChangedStats, but the line above (98). The check will always yield True, even if there is no change.

gperonato: Line 98 compares two dicts, which to me seems ok

gperonato: Ah I see now, there's a problem with the parentheses

data-integration/getJRC_GEOPP_DB_csv.py
melhk: Ok thanks!
It would be cleaner clearer and however not to depend on prints to control the logic of the code, but to put it directly in the control structures.

data-integration/getHotMaps_raster.py
melhk: Yes that's what I meant, with the "[...]"!

neodark: Exact that's a dangerous statement :wink: 
`if [dp["resources"][r]["stats"] != new_dp["resources"][r]["stats"]]` or `if dp["resources"][r]["stats"] != new_dp["resources"][r]["stats"]` differ completely.
Good catch :+1:

melhk:
There are still conflicting files with the master branch. Can you solve them please?

neodark:
To all: 

* When someone submits a Pull Request. The reviewer has to merge it. **This makes sure we have the second pass of check in order to patch/review code.**
* Also before merging, **always make sure there are no conflicts with the master branch**.
* I know this is painful to all but don't forget: **unit tests**. This ensures code quality on the long term.
* **Thinks twice before merging to master**: Never merge to master a branch with conflicts. Someone else pulling master will have issues.

Thanks to all for your work!

gperonato:
Dear @neodark,
Thank you for the kind reminder.
Of course I am aware of this and it was clearly my mistake. As agreed already with @melhk, we decided not to revert for now the merge.
Looking forward to continuing a great collaboration

neodark:
Hi @gperonato thanks for your comment. I was not aware of your discussions. It was just a general remark.
Thanks to you and all for the great job on this project. :+1:

neodark:
Hello everyone
unfortunately we had to rewrite history and remove this merge commit because it contained some important comments that contained some unresolved points and added some linter problem on a security issue.

- The history as therefore been re-written on GItHub
- You therefore need to delete your master branch
$ git checkout -b tempbranch # switch to a temporary branch
$ git branch -D master # hard delete master branch

- You have to clone again the new master branch
$ git checkout master # get the new master branch
$ git branch -D tempbranch # delete temporary branch

- Create a new development branch from the new master branch
$ git checkout -b mynewdevelbranch

- Cherry-pick commits from your previous branch to this new development branch (if any work in progress)
$ git cherry-pick  -m 2 <commit hash from old working branch>

- Delete your old working branch and work on this new development branch
$ git branch -D oldworkingbranch

And you can create PR again on GitHub

Please do a new Pull Request for the branch that was reverted and we'll go over the remaining unresolved points prior to merging it again.

PR #188, opened 2021-02-25
adding docstring

### LINKED ISSUE #175: check if all functions have docstring

cedricserugendo:
add docstring on functions


api/app/endpoints/wms.py
```diff
@@ -65,14 +83,15 @@ def parse_size(params) -> Size:
...
     if (height * width) > current_app.config["WMS"]["MAX_SIZE"]:
-        abort(400, "Total size is bigger than the maximaml allowed size")
+        abort(400, "Total size is bigger than the maximal allowed size")
     return Size(width=width, height=height)
 
 
 Position = namedtuple("Position", ("x", "y"))
 
 
 def parse_position(params) -> Size:
```
melhk, line 93: The type of the variable returned is not matching the type the function expects??

melhk: Why does this function return an other type than the one expected?

melhk (commented):
1 comment to check

api/app/endpoints/wms.py
cedricserugendo: This is a mistake, the function should return the namedtuple `Position` (see lines 293-296)
I'll fix it

melhk (approved):

PR #204, opened 2021-04-19
warping if srs has changed

### LINKED ISSUE #203: Reprojection with gdalwarp does not work in Docker
This affects raster datasets that need to be converted to EPSG:3035

gperonato:
I realized that `gdaltransform` was not changing the projection: `gdal_warp` is needed in this case (notably for `ds_id 31`).
I also:
- changed the compression algorithm (to avoid huge files)
- fixed https://github.com/enermaps/enermaps/issues/203
Sorry for these late fixes.

Reprojection could be very slow, depending on Docker settings. I hence changed the sample raster dataset to 43, which does not require reprojection.

### DIFF
--- a/data-integration/utilities.py
+++ b/data-integration/utilities.py
@@ -305,21 +305,4 @@ def get_ld_json(url: str) -> dict:
         "".join(soup.find("script", {"type": "application/ld+json"}).contents)
     )
 
-def extractZip(source, target):
-    """
-    Extract zip.
-    Parameters
-    ----------
-    source : string or path-like object
-    target : string
-    Returns
-    -------
-    List of strings: extracted file path
-    """
-    # Get the file names of extracted files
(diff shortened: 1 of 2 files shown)


melhk:
Hello @gperonato, can you pull the latest changes from the master into your branch please?

data-integration/utilities.py
```diff
@@ -304,3 +304,22 @@ def get_ld_json(url: str) -> dict:
...
+    Parameters
+    ----------
+    source : string or path-like object
+    target : string
+    Returns
+    -------
+    List of strings: extracted file path
+    """
+    # Get the file names of extracted files
+    zip_ref = zipfile.ZipFile(source, "r")
```
melhk: It seems to me that this is not used

gperonato: You are right, sorry. It has been removed now.

melhk (approved):

PR #210, opened 2021-04-29
Correction of args errors and CM empty

melhk:


cm/base/BaseCM/cm_output.py
```diff
@@ -74,19 +74,17 @@ class CMOutput(Schema):
...
+    # data structure
+    return output_schema.load(data=output)
 
 
 def output_raster(raster_name, raster_fd):
     """Add a raster to the api."""
     files = {"file": (raster_name, raster_fd, "image/tiff")}
     try:
-        resp = requests.post(API_URL + "api/geofile/", files=files)
+        API_URL = os.environ.get("API_URL")
```
neodark: Why is this in the function and not a global data from the application ?

cm/example_empty/worker.py
```diff
@@ -1,17 +1,18 @@
...
 @app.task(base=cm_base.CMBase, bind=True, schema_path=schema_path)
-def new_cm(self, file: str, params: dict):
-    """New cm to test the integration work flow"""
-    status = process(file)
-    return status
+def cm_empty(self, selection: dict, rasters: list, params: dict):
+    ret = dict()
+    ret["graphs"] = {}
+    ret["geofiles"] = {}
+    ret["values"] = {}
```
neodark, line 14: as it doesn't add extra information we can have a one liner here:
`ret = {"graphs":{}, "geofiles":{}, "values":{}}`

melhk: I find it easier for the future "implementers" to understand what to do if it is already decomposed and not merged in one line.

cm/base/BaseCM/cm_output.py
melhk: I don't know, I have only corrected a "local bug" for now.....

melhk: I don't know, I have only corrected a bug "locally" for now...

cm/example_empty/worker.py
neodark: Agreed due to the future coders level.

neodark (approved):
Looks good to me

PR #211, opened 2021-04-29
[code cleanup] Pre-commit hooks

neodark:
This PR will help automate the programmer's ability to follow python coding guidelines.

Once pre-commit installed (as mentioned in the README file), `git commit` will locally trigger pre-commit to launch a pass of
- black
- isort
- flake8
- bandit

The code will get modified automatically if it doesn't follow the guidelines or suggestions to change will pop up and the commit will not get accepted.
Once all rules are followed the commit will pass and will be integrated to the local repository.
The user can done push to the remote repository and create a PR.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,0 +1,4 @@
+black
+bandit
+flake8
+pre-commit
--- a/.flake8
+++ b/.flake8
@@ -1,0 +1,4 @@
+[flake8]
+max-line-length = 88
+select = B,C,E,F,W,T4,B9,B950
+ignore = E501, W503, E203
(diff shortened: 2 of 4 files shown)


requirements.txt
```diff
@@ -0,0 +1,4 @@
+black
+bandit
+flake8
+pre-commit
```
melhk, line 4: Shouldn't we specify the same versions that are given in the file linter-requirements-txt?

neodark: I prefer not. Actually we should be using the latest ones. I think this is something that should be improved in `linter-requirements.txt` too

melhk (approved):

PR #212, opened 2021-04-30
Location of data volume from data-integration

gperonato:
This might help with retrieving the raster files from other docker services.
You can mount the `data` volume in another service and there you should find the raster files that have been integrated. The path will be: `data/{ds_id}/{fid}` where `fid` is the name of the GeoTiff file.

### DIFF
--- a/data-integration/getHotMaps_raster.py
+++ b/data-integration/getHotMaps_raster.py
@@ -11,11 +11,11 @@ import argparse
 import json
 import logging
 import os
+import shutil
 import sys
 
 import frictionless
 import pandas as pd
-
 import utilities
 
 # Constants
@@ -93,7 +93,7 @@ def get(repository: str, dp: frictionless.package.Package, isForced: bool = Fals
             }
             rasters.append(raster)
             # check statistics for each resource
-            if dp != None and "stats" in new_dp["resources"][resource_idx]:
+            if dp is not None and "stats" in new_dp["resources"][resource_idx]:
                 if (
                     dp["resources"][resource_idx]["stats"]
                     != new_dp["resources"][resource_idx]["stats"]
@@ -101,7 +101,7 @@ def get(repository: str, dp: frictionless.package.Package, isForced: bool = Fals
                     isChangedStats = True
     rasters = pd.DataFrame(rasters)
 
-    if dp != None:  # Existing dataset
+    if dp is not None:  # Existing dataset
         # check stats
         isChangedVersion = dp["version"] != new_dp["version"]
         if isChangedStats or isChangedVersion:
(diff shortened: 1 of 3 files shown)


data-integration/getHotMaps_raster.py
```diff
@@ -93,15 +93,15 @@ def get(repository: str, dp: frictionless.package.Package, isForced: bool = Fals
...
+            if dp is not None and "stats" in new_dp["resources"][resource_idx]:
                 if (
                     dp["resources"][resource_idx]["stats"]
                     != new_dp["resources"][resource_idx]["stats"]
                 ):
                     isChangedStats = True
     rasters = pd.DataFrame(rasters)
 
-    if dp != None:  # Existing dataset
+    if dp is not None:  # Existing dataset
```
gcmalloc, line 104: writing

```python
if dp: 
```
is concidered more pythonic.

gperonato: dp is never a boolean, I don't think it would work in this case

melhk: Yes it works, because you are testing for identity with "None", not if dp has the value "True" or "False".
"True" is not "None", but "False" is not "None" either.

gperonato: Not in this case:
```
dp = {}
if dp:
	print("test1")


if dp is not None:
	print("test2")
```
Anyway, I don't think this is the point of this commit. And the non-pythonic way was accepted in the previous merge. Here the change was from `!=` to `is not`, because the previous version failed the precommit.

melhk (approved):

data-integration/getHotMaps_raster.py
melhk: No, this is wrong and you should change it.

Your example doesn't print "test1", because if you test an empty dictionary, with no keys, it returns "False".
```
>>> dp1 = {}
>>> dp2 = {"version": None}
>>>
>>> if dp1 is not None: print("test1")
... 
prints test1
>>> if dp1: print("test2")
...
prints nothing
>>> if dp2 is not None: print("test3")
...
prints test3
>>> if dp2: print("test4")
...
prints test4
>>>
```

In the following lines of your code, you make the test `isChangedVersion = dp["version"] != new_dp["version"]` which requires that you have the key "version" existing in your dictionary.

But if you check your dictionary like that `if dp is not None`, you may encounter a key error:
```
>>> dp1["version"] == dp2["version"]
  File "<stdin>", line 1, in <module>
KeyError: 'version'
```

that can't happen if you check your dictionary with `if dp`.

gperonato: Ok, I will correct it. Can you please open an issue for that?

neodark: Here maybe some piece of explanation for all: `https://www.python.org/dev/peps/pep-0008/` under `programming recommendations`

```
Comparisons to singletons like None should always be done with `is` or `is not`, never the equality operators.

Also, beware of writing `if x` when you really mean `if x is not None` 
-- e.g. when testing whether a variable or argument that defaults to `None` was set to some other value. 
The other value might have a type (such as a container) that could be false in a boolean context!
```

melhk: > 
> 
> Ok, I will correct it. Can you please open an issue for that?

Why open an issue for that ?

cedricserugendo (approved):

PR #224, opened 2021-05-29
add method that validate area selection before running the CM

### LINKED ISSUE #221: Selection problem at the LAU level if no pixel have data. Problem identified, but lack of implementation.

### LINKED ISSUE #229: Add in the documentation of the CM that the case without data and with too much data must be managed

cedricserugendo:


cm/base/BaseCM/cm_input.py
```diff
@@ -1,8 +1,80 @@
...
+
+
+def validate_selection(
+    selection: dict,
+    raster: str,
+):
+    """
+    Find the number of count for given a raster and a area selection in order to define if the
+    selection is valid or not.
+
```
melhk, line 37: What is a count?

cm/base/BaseCM/cm_input.py
```diff
@@ -1,8 +1,80 @@
...
     raster_dir = os.path.join(os.environ["UPLOAD_DIR"], "raster")
     raster_path = os.path.join(raster_dir, raster_name, "raster.tiff")
     return raster_path
+
+
+class StatNotComputeError(Exception):
+    """Exception thrown for a non-existing calculation module"""
+
+    pass
+
```
melhk, line 26: Why is there an unimplemented function?

cm/base/BaseCM/cm_input.py
```diff
@@ -1,8 +1,80 @@
...
+    """
+    Find the number of count for given a raster and a area selection in order to define if the
+    selection is valid or not.
+
+    Inputs:
+        * raster : path to the raster selected from the frontend.
+        * selection : area selection made from the frontend
+    Outputs:
+        * selection_valid : boolean that define if raster selection is valid.
+        * response : response return to the front end when are selected is not valid.
```
melhk: Please check the grammar

cm/base/BaseCM/cm_input.py
cedricserugendo: it is a class used to raise error

cm/base/BaseCM/cm_input.py
cedricserugendo: The count is a non-null pixel of a raster.. I add this information in the docstring

cm/base/BaseCM/cm_input.py
cedricserugendo: sorry, the docstring is fixed

cm/base/BaseCM/cm_input.py
```diff
@@ -31,13 +31,14 @@ def validate_selection(
     """
     Find the number of count for given a raster and a area selection in order to define if the
```
melhk: Grammar to check

cedricserugendo: the correction is made

melhk:
Please make sure that all tests pass

melhk:
It's still failing :(

neodark:
The issue here is mentioned in the output of the linter: `Do not use bare except, specify exception instead (E722)`

https://www.flake8rules.com/rules/E722.html and https://www.python.org/dev/peps/pep-0008/#programming-recommendations gives you more explanation.

It's a good practice to specify the error you're catching in general or catch all exceptions with `Exception` instead of a bare except.

Reason:

 `A bare except: clause will catch SystemExit and KeyboardInterrupt exceptions, making it harder to interrupt a program with Control-C, and can disguise other problems. If you want to catch all exceptions that signal program errors, use except Exception: (bare except is equivalent to except BaseException:)`

cm/base/BaseCM/cm_input.py
```diff
@@ -1,8 +1,87 @@
...
 
 def get_raster_path(raster_name):
     """Returns the path to the raster file based on the raster name."""
     raster_dir = os.path.join(os.environ["UPLOAD_DIR"], "raster")
     raster_path = os.path.join(raster_dir, raster_name, "raster.tiff")
     return raster_path
+
+
+class StatNotComputeError(Exception):
+    """Exception thrown for a non-existing calculation module"""
```
melhk: The description does not seem to match the function name?

cm/base/BaseCM/cm_input.py
```diff
@@ -1,8 +1,87 @@
...
+    if count <= 0:
+        response["values"] = {"No count found:": 0}
+    else:
+        if max_count is not None and count > max_count:
+            response["values"] = {
+                "Too many count found (max." + str(max_count) + "):": str(count)
+            }
+        else:
+            selection_valid = True
+    validate(response)
```
melhk, line 86: Why do you try to validate an empty response? This function should only check the input of the calculation module.

cedricserugendo: I validate the answer because in case none or too many counts are found, there is an error message that is returned thanks to the value linked to the "values" key

cm/base/BaseCM/cm_input.py
cedricserugendo: that is correct, my bad
I will fix it

melhk (approved):

PR #239, opened 2021-06-10
Adding new data-integration pipelines

gperonato:


.pre-commit-config.yaml
```diff
@@ -25,7 +25,7 @@ repos:
...
     hooks:
       - id: flake8
-  - repo: https://github.com/PyCQA/bandit
-    rev: 1.6.2
-    hooks:
-      - id: bandit
+  # - repo: https://github.com/PyCQA/bandit
+  #   rev: 1.6.2
+  #   hooks:
+  #     - id: bandit
```
gcmalloc: why is that commented out ?

data-integration/Dockerfile
```diff
@@ -7,13 +7,16 @@ RUN apt-get update &&\
...
     gdal-bin \
+    proj-bin \
     libgdal-dev &&\
     rm -rf /var/cache/apt/archives/
 
 
 ENV CPLUS_INCLUDE_PATH=/usr/include/gdal
 ENV C_INCLUDE_PATH=/usr/include/gdal
 
+COPY .cdsapirc /root/.cdsapirc
```
gcmalloc: from the build stack, it seems that this file is not in the directory and thus the dockerfile cannot be built, any way to accept the term of service of copernicus another way ?

data-integration/admin_units.py
```diff
@@ -110,12 +113,39 @@ def get(
...
+    # Upload dataset record
     dataset = pd.DataFrame([{"ds_id": DS_ID}])
     utilities.toPostgreSQL(
         dataset, DB_URL, schema="datasets",
     )
+    # Upload spatial records
+    admin_units["ds_id"] = DS_ID
     utilities.toPostGIS(admin_units, DB_URL)
+    # Upload empty data records
+    data = admin_units.loc[:, ["ds_id", "fid"]].copy()
```
gcmalloc: Everything is data, thus this is not really a good variable name. Maybe renaming it feature would be better ?

data-integration/admin_units.py
```diff
@@ -110,12 +113,39 @@ def get(
...
+    )
+
+
+if __name__ == "__main__":
+    argv = sys.argv
+    if "--force" in argv:
+        isForced = True
+    else:
+        isForced = False
+    if utilities.datasetExists(DS_ID, DB_URL):
```
gcmalloc, line 128: In this particular case it is not so important, but usually this will create toctou race condition.
There is a difference between the time of check and time of use (here the insertion) where another data pipeline could have run and inserted the same dataset. 
Instead you could insert anyway and catch the raised error. This mantra is refered as "don't ask for permission, ask for forgiveness".

data-integration/admin_units.py
```diff
@@ -110,12 +113,39 @@ def get(
...
+        if isForced:
```
gcmalloc, line 129: I think you want an upsert operation to the database here.

data-integration/getBuildingHeight.py
```diff
@@ -0,0 +1,190 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
```
gcmalloc: in python3, the default encoding is utf-8, so you don't need those anymore, see https://stackoverflow.com/questions/14083111/should-i-use-encoding-declaration-in-python-3.

data-integration/getBuildingHeight.py
```diff
@@ -0,0 +1,190 @@
...
+"""
+Prepare the Building Height dataset for EnerMaps.
+
+Note that the  files must be downloaded from Copernicus (requires log-in)
+and extracted in the data/31 directory.
+This script expects that the original zip file from Copernicus is extracted
+in multiple zip files.
+Created on Wed June 02 12:50:00 2021
```
gcmalloc: I'm not sure how the creation date helps given that we already have git tracking all modification of that file.

data-integration/getBuildingHeight.py
```diff
@@ -0,0 +1,190 @@
...
+@author: giuseppeperonato
+"""
+
+import glob
+import json
+import logging
+import os
+import pathlib
+import shutil
+import sys
```
gcmalloc, line 21: the import order looks odd, could you run isort on it ?

data-integration/getBuildingHeight.py
```diff
@@ -0,0 +1,190 @@
...
+DB_PASSWORD = os.environ.get("DB_PASSWORD")
+DB_DB = os.environ.get("DB_DB")
+
+DB_URL = "postgresql://{DB_USER}:{DB_PASSWORD}@{DB_HOST}:{DB_PORT}/{DB_DB}".format(
+    DB_HOST=DB_HOST,
+    DB_PORT=DB_PORT,
+    DB_USER=DB_USER,
+    DB_PASSWORD=DB_PASSWORD,
+    DB_DB=DB_DB,
+)
```
gcmalloc: I think those are used on every file below, could you move those to utilities as a variable and then do 
`DB_URL = utilities.DB_URL` ?

data-integration/admin_units.py
```diff
@@ -110,12 +113,39 @@ def get(
...
+    data["value"] = 0
+    data["variable"] = ""
+    utilities.toPostgreSQL(
+        data, DB_URL, schema="data",
+    )
+
+
+if __name__ == "__main__":
+    argv = sys.argv
+    if "--force" in argv:
```
gcmalloc: could you use argparse here instead ?

data-integration/getBuildingHeight.py
```diff
@@ -0,0 +1,190 @@
...
+    xmax = xmin + width * xpixel
+    ymin = ymax + height * ypixel
+
+    return box(xmin, ymin, xmax, ymax)
+
+
+def convertZip(directory: str):
+    """Convert files downloaded from Copernicus."""
+    files_list = glob.glob(os.path.join(directory, "*.zip"))
+    if len(files_list) > 0:
```
gcmalloc: you can write:

```
if files_list:
```
I usually try to have a plural for collections, so I would rather write
```
    zip_files = glob.glob(os.path.join(directory, "*.zip"))
    if zip_files:
        ...
```

data-integration/getBuildingHeight.py
```diff
@@ -0,0 +1,190 @@
...
+        logging.info("Extracting zip files")
+        for zipfile in files_list:
+            extract_dir = os.path.join(
+                os.path.dirname(zipfile), pathlib.Path(zipfile).stem
```
gcmalloc: usage of pathlib here is a bit odd given that you join file with os.path.join, maybe using os.path.basename instead of the .stem would make more sense ?

data-integration/getBuildingHeight.py
```diff
@@ -0,0 +1,190 @@
...
+
+    return box(xmin, ymin, xmax, ymax)
+
+
+def convertZip(directory: str):
+    """Convert files downloaded from Copernicus."""
+    files_list = glob.glob(os.path.join(directory, "*.zip"))
+    if len(files_list) > 0:
+        logging.info("Extracting zip files")
+        for zipfile in files_list:
```
gcmalloc: Wth the proposal above,  you could do here
```
for zip_file in zip_files:
...
```

data-integration/getBuildingHeight.py
```diff
@@ -0,0 +1,190 @@
...
+            extract_dir = os.path.join(
+                os.path.dirname(zipfile), pathlib.Path(zipfile).stem
+            )
+            extracted = utilities.extractZip(zipfile, extract_dir)
+            source_file = [x for x in extracted if x.endswith("tif")][0]
```
gcmalloc: this will crash if we have no tif in the file, maybe broke it into two part:
```
            tif_files = [x for x in extracted if x.endswith("tif")
            if not tif_files:
                log.warning("no tiff file found in zipfile")           
                 continue
             ....
```

data-integration/getBuildingHeight.py
```diff
@@ -0,0 +1,190 @@
...
+
+            logging.info(source_file)
+
+            dest_file = os.path.join(
+                os.path.dirname(extract_dir), os.path.basename(extract_dir) + ".tif",
```
gcmalloc: I'm not sure I understand what we are doing here
os.path.split(extract_dir) will return dirname and basename (see https://docs.python.org/3/library/os.path.html#os.path.split) so joining them together will return extract_dir.
As I understand this is the equivalent of doing extract_dir + '.tif', is there something I don't understand ?

data-integration/getBuildingHeight.py
```diff
@@ -0,0 +1,190 @@
...
+            )
+            os.system(
```
gcmalloc: You will have some nasty surprise if source file or dest file contains space. 
Use subprocess.check_output for example instead, without the shell set to true.

data-integration/getBuildingHeight.py
```diff
@@ -0,0 +1,190 @@
...
+    extents = []
+    for file in files_list:
+        logging.info(file)
+        src_ds = gdal.Open(file)
+        prj = src_ds.GetProjection()
+        srs = osr.SpatialReference(wkt=prj)
+        source_crs = CRS.from_epsg(srs.GetAttrValue("authority", 1))
+
+        try:
+            assert source_crs.to_string() == "EPSG:3035"
```
gcmalloc: Why not just testing with a comparison here ?

```
if source_crs.to_string() != "EPSG:3035":
   logging.error("input files must be in")
```

data-integration/getBuildingHeight.py
```diff
@@ -0,0 +1,190 @@
...
+            "start_at",
+            "fields",
+            "variable",
+            "value",
+            "ds_id",
+            "fid",
+            "dt",
+            "z",
+            "israster",
+            "unit",
```
gcmalloc: If those are coming directly from the database, it would be better to put them in a central place such as utilities.

data-integration/getBuildingHeight.py
```diff
@@ -0,0 +1,190 @@
...
+        ]
+    )
+    enermaps_data["fid"] = fids
+
+    spatial = gpd.GeoDataFrame(geometry=extents, crs="EPSG:3035",)
```
gcmalloc: the projection is reused in lots of places, maybe it needs to be put in a variable ?

data-integration/getBuildingHeight.py
```diff
@@ -0,0 +1,190 @@
...
+    spatial["fid"] = fids
+
+    return enermaps_data, spatial
+
+
+if __name__ == "__main__":
+    argv = sys.argv
```
gcmalloc: similar remark about using argparse, if all your script are using the same set of argument you can have a single parser created in utilities and use that one.

data-integration/getBuildingHeight.py
```diff
@@ -0,0 +1,190 @@
...
+    datasets = pd.read_csv("datasets.csv", engine="python", index_col=[0])
```
gcmalloc: here the encoding doesn't seems to be explicitely specified, maybe using utf-8 for the dataset list would be nice ?

data-integration/getBuildingHeight.py
```diff
@@ -0,0 +1,190 @@
...
+    if "--force" in argv:
+        isForced = True
+    else:
+        isForced = False
+
+    directory = "./data/{}".format(ds_id)
+
+    if (
+        os.path.exists(directory)
+        and os.path.isdir(directory)
```
gcmalloc: os.path.isdir(directory) will return True if the file exist and is a directory, so you don't need the first check.

data-integration/getBuildingHeight.py
```diff
@@ -0,0 +1,190 @@
...
+        # Create data table
+        data["ds_id"] = ds_id
+        utilities.toPostgreSQL(
+            data, DB_URL, schema="data",
+        )
+        # Create spatial table
+        spatial["ds_id"] = ds_id
+        utilities.toPostGIS(
+            spatial, DB_URL, schema="spatial",
+        )
```
gcmalloc: When interacting with a database, you try to do everything that should fail/succeed in one transaction, so here inserting in both three table as a single transaction would be the correct approach.
Else if the script is crashing between any of those query you will end up with a partially inserted dataset.

data-integration/getBuildingHeight.py
```diff
@@ -0,0 +1,190 @@
...
+    else:
+        logging.error(
+            "The {} directory must exist and contain {} files from Copernicus.".format(
+                directory, N_FILES
```
gcmalloc: Here pyflakes would give you a warning as logging function already use a format internally.

```
       logging.error(
            "The %s directory must exist and contain %s files from Copernicus.",
                directory, N_FILES)

File: data-integration/getEEA.py [hunk 25]
```diff
@@ -0,0 +1,208 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+"""
+Created on Mon Mar 8 11:51:38 2021
```
gcmalloc: Missing comment about the module.

data-integration/getEEA.py
```diff
@@ -0,0 +1,208 @@
...
+import pandas as pd
+import utilities
+import frictionless
+from pandas_datapackage_reader import read_datapackage
+import json
+import os
+import sys
+from datetime import datetime
+import requests
+import zipfile
```
gcmalloc: the sorting of import is wrong, it should follow the canonical order:

    Standard library imports.
    Related third party imports.
    Local application/library specific imports.

isort will do that for you.

data-integration/getEEA.py
```diff
@@ -0,0 +1,208 @@
...
+    chunk_size : integer, optional
+        The default is 128.
+
+    Returns
+    -------
+    None.
+
+    """
+    r = requests.get(url, allow_redirects=True, stream=True,timeout=timeout)
+    if len(append_path) > 0:
```
gcmalloc: You can replace this by
```
if append_path:
...
```

data-integration/getEEA.py
```diff
@@ -0,0 +1,208 @@
...
+    filepath, last_modified  = extractZip("{}.zip".format(path),path)[0]
+
+    # Metadata to retrieve from data
+    last_modified = datetime(*last_modified).strftime("%Y-%m-%d %H:%M:%S")
+    # last_modified = datetime.fromtimestamp(utilities.creation_date(filepath)).strftime('%Y-%m-%d-%H:%M')
+    # year = int(filepath.split("/")[-1].split("-")[0])
+    
+    # Rename the data, so that the path is the same regardless of the version
+    os.rename(filepath,os.path.join(path,"data.csv"))
+    os.remove("{}.zip".format(path))
```
gcmalloc: nit. `path + ".zip"` ?

data-integration/getEEA.py
```diff
@@ -0,0 +1,208 @@
...
+                                           )
+    new_dp.resources[0]["path"] = os.path.join(path,"data.csv")
+    # Add date
+    new_dp["last_modified"] = last_modified
+    # Add groupChar
+    new_dp.resources[0].schema.get_field("ValueNumeric").group_char =","
+    # new_dp.to_json("{}/datapackage.json".format(path))
+    
+    # Logic for update
+    if dp != None:
```
gcmalloc: I think this already was said in another review, None is a singleton in python so you should us `is not` for the comparison.

data-integration/getEEA.py
```diff
@@ -0,0 +1,208 @@
...
+        # check stats
+        ChangedStats = dp["resources"][0]["stats"] != new_dp["resources"][0]["stats"]
+        ChangedDate = dp["last_modified"] != new_dp["last_modified"]
+
+        if ChangedStats or ChangedDate:
+            print("Data has changed")
```
gcmalloc: What is the reason for using print here when the other files are using logging ?

data-integration/getEEA.py
```diff
@@ -0,0 +1,208 @@
...
+    if "--force" in argv:
+        Force = True
+    else:
+        Force = True
+
+    if not os.path.exists("data"):
+        os.mkdir("data")
+    path = os.path.join("data",str(ds_id))
+    if not os.path.exists(path):
+        os.mkdir(os.path.join(path))
```
gcmalloc: here use this instead of doing the check, create move one directory deeper:

```
 os.makedirs(path, exist_ok=True)
```

This will create the directory tree if not yet present and do nothing if the directory tree structure is already present.

data-integration/getEEA.py
```diff
@@ -0,0 +1,208 @@
...
+        
+    dp = utilities.getDataPackage(ds_id, "postgresql://test:example@{host}:{port}/dataset".format(host=host,port=port))
```
gcmalloc: What is the reason for not using the DB_URL here ?

data-integration/getENER.py
```diff
@@ -0,0 +1,181 @@
...
+}
+
+FILES = {
+    "WP3_DataAnnex_FinalEnergy_ForPublication_201607.xlsx": "Final Energy",
+    "WP3_DataAnnex_PrimaryEnergy_ForPublication_201607.xlsx": "Primary Energy",
+    "WP3_DataAnnex_UsefulEnergy_ForPublication_201607.xlsx": "Useful Energy",
+}
+
+
+# In Docker
```
gcmalloc: why only in docker ? you can set the environment variable in any environment you want. For example by having a file containing
```
export DB_HOST=localhost
export DB_PORT=5432
...
```
then source the file and you can reuse all the script outside of docker. 
Similar remark concerning the repetition of the DB_URL definition.

gcmalloc (changes requested):
Review is way too big to be reviews at once, I started with the start. 
In general, could you run black and isort ?

.pre-commit-config.yaml
gperonato: There was a problem with this:
https://github.com/enermaps/enermaps/blob/358a67cb046b727a0665e5687ef2adb3ba139b79/data-integration/utilities.py#L83

data-integration/Dockerfile
gperonato: No, this is the only way.
Any suggestion to make the build stack working also without the .cdsapirc file?

data-integration/admin_units.py
gperonato: I see, but I don't think in this case is a problem. Data integration pipelines will be run sequentially.

data-integration/admin_units.py
gperonato: Sorry I don't understand

data-integration/getBuildingHeight.py
gperonato: The use of Pathlib here is to get the file name without extension.

data-integration/getBuildingHeight.py
gperonato: I see, but I don't think here it is a major problem.

data-integration/Dockerfile
gcmalloc: Maybe mount it as a volume instead, so that it doesn't end up in the final image ?

data-integration/admin_units.py
gcmalloc: It's a similar remark as the one above the race condition, usually you try to bundle database operation that should be done in one operation into a single transaction. If you start running multiple pipeline in parallel or if something else is accessing the database, this will be a problem.

data-integration/getBuildingHeight.py
gcmalloc: so the equivalent using the os module would be:

```python
os.path.splitext(zipfile)[0]
```

.pre-commit-config.yaml
gcmalloc: You can use popen.check_output with shell set to false as an alternative. You can also ignore bandit check for a particular instance by putting the `# nosec` comment (see https://stackoverflow.com/questions/52596576/what-is-the-way-to-ignore-skip-some-issues-from-python-bandit-security-issues-re).

data-integration/Dockerfile
gperonato: How would that work? I have tried but it doesn't work. This should be a mounted volume?

gperonato: Ideally, I would need a folder containing the API file that maps to /root in the container.

.pre-commit-config.yaml
gperonato: The `# no sec` will likely conflict with black...

gperonato: I was wrong, it works

data-integration/Dockerfile
gperonato: To be solved in next PR, these pipelines do not need CDS

data-integration/admin_units.py
```diff
@@ -1,41 +1,38 @@
...
-    DB_HOST=DB_HOST,
-    DB_PORT=DB_PORT,
-    DB_USER=DB_USER,
-    DB_PASSWORD=DB_PASSWORD,
-    DB_DB=DB_DB,
+    DB_HOST="localhost",
+    DB_PORT=5433,
+    DB_USER="test",
+    DB_PASSWORD="example",
+    DB_DB="dataset",
```
gcmalloc: Maybe having this using the same approach as the other file would be better ?

gcmalloc (approved):

data-integration/admin_units.py
neodark: Shouldn't this come from the `utilities` you implemented in this PR ?

PR #242, opened 2021-06-15
Create postgres user from .env - mitigate default user/password

neodark:
The user/password isn't created in the sql script.

Create postgres user from .env - mitigate default user/password on file

Postgres has a default test user defined

### DIFF
--- a/.env
+++ b/.env
@@ -4,7 +4,7 @@ CELERY_RESULT_BACKEND=redis://redis
 API_URL=http://api
 DB_HOST=db
 DB_PORT=5432
-DB_USER=test
+DB_USER=api
 DB_PASSWORD=example
 DB_DB=dataset
 
--- a/.env
+++ b/.env
@@ -4,7 +4,7 @@ CELERY_RESULT_BACKEND=redis://redis
 API_URL=http://api
 DB_HOST=db
 DB_PORT=5432
-DB_USER=api
+DB_USER=test
 DB_PASSWORD=example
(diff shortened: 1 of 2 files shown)


neodark:
@gperonato : this improves somehow on our private discussion we had regarding security, default users and passwords.

We have another user (api) that is different from Postgres default user.
And the user/password is pulled from .env instead of being hardcoded in the script, so we could easily change on production at a future step.

Note: Postgres has a terrible authentication mechanism with default users etc.

.env
```diff
@@ -4,7 +4,7 @@ CELERY_RESULT_BACKEND=redis://redis
 API_URL=http://api
 DB_HOST=db
 DB_PORT=5432
-DB_USER=test
+DB_USER=api
```
gperonato: I would use another name, as this is not the user used by postgrest

melhk: I don't understand what you mean by that?

gperonato: This is confusing with the role used by PostGREST here
https://github.com/enermaps/enermaps/blob/378e04420e81317397e7fae3cfbbfa8cf8177db3/db/init/postgrest.sql#L6

neodark: 1. PostgREST is still not on master. This was just to comply with the current setup.
2. The username here could be whatever is preferred. Ok so I'll set it back to the previous user name "test"

neodark: Done. pushed.

melhk (approved):

PR #252, opened 2021-06-25
create empty .env-local file

gperonato:
This is a workaround to have a local env file (not tracked by git) and be able to run CI in GitHub.
Inspired by this: https://github.com/docker/compose/pull/3955#issuecomment-847646545

### DIFF
--- a/.github/workflows/pr_to_master.yml
+++ b/.github/workflows/pr_to_master.yml
@@ -70,5 +70,7 @@ jobs:
     runs-on: 'ubuntu-18.04'
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


chrisdlangton:
Support :+1: 

Why? go read why:
https://github.com/docker/compose/pull/3955#issuecomment-870198643

gperonato:
Provisionally in draft, as this seems not resolutive.
See thread in Mattermost.

gperonato:
The strategy is different.
We keep secret keys in separate files, instead of a single .env.
Empty files are to be created by Github actions

PR #257, opened 2021-07-12
Post raster

cedricserugendo:
add correction in the `output_raster` function

### DIFF
--- a/cm/base/BaseCM/cm_output.py
+++ b/cm/base/BaseCM/cm_output.py
@@ -85,7 +85,7 @@ def output_raster(raster_name, raster_fd):
     """Add a raster to the api."""
     files = {"file": (raster_name, raster_fd, "image/tiff")}
     try:
-        resp = requests.post(API_URL + "/geofile/", files=files)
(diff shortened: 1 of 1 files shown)


cm/base/BaseCM/cm_output.py
```diff
@@ -85,7 +85,7 @@ def output_raster(raster_name, raster_fd):
     """Add a raster to the api."""
     files = {"file": (raster_name, raster_fd, "image/tiff")}
     try:
-        resp = requests.post(API_URL + "/geofile/", files=files)
+        resp = requests.post(API_URL + "/api/geofile/", files=files)
```
melhk, line 88: Why do you want to write something like "api/api/geofile" ?

cedricserugendo: `API_URL`  is the base URL 
for the moment it returns `http://api` but at the end it should return something like `https://enermaps.eu/` 
then to post a geofile, we need to go to `/api/geofile` endpoint

melhk (approved):

PR #269, opened 2021-08-09
Integration pipeline for  esm eudem

gperonato:


neodark:
Please rebase on master this PR.
Take into account that the db stuff has been split in a separate docker-compose for more clarity. The README in data-integration with this PR needs to be updated accordingly and re-tested if it still works. (There's now a `-f ../docker-compose-db.yml` before the python script to run)

neodark:
In: https://github.com/enermaps/enermaps/pull/271/files they were also modifications in :

- datasets.csv
- utilities.py

But not here. Did you forget to update those files in this PR or is it fine like this ?

gperonato:
Yes, good point. They were not modified here, but the master should take into account the changes from all pull requests.
Should I merge here the changes in the other PR so that there are no conflicts afterward? Or create a new branch based on the three current branches I want to merge?

neodark:
Actually `getElectricity.py` is used in PR 271 and here `getESM-EUDEM.py` is not mentioned in the csv file, so I guess it will never call the script even if we merge that PR first.

My initial thought was to merge 269, 270, 271 in that order, but if you prefer another order, it's fine with me, just let me know.
But again you will have the changes from 271 with the csv updated with `getElectricity.py` but still `getESM-EUDEM.py` will never be injected in master as it's not mentioned in any of the 3 PRs.

I propose to merge PR 271 which seems more complete, so you can rebase this one on master and see what is missing, then fix if need and then we merge this one and we can do the same process on PR 270 at last.

Does this sound fine? Let me know what you prefer.

gperonato:
Yes, that would also work. You can go with PR271, then I'll have to rebase the other ones.
As an alternative, #272 contains all three PRs (but maybe is less transparent).

Since there are still a lot of pipelines that need to be integrated, please let me know which way you prefer.

neodark:
Actually #272 would make it more difficult to spot, buut I understand why you proposed it as an alternative. We can do that for the future pipelines.
As I merged #271 I propose we go one by one for #269 and #270 and delete #272.
There's no problem creating multiple PR. We can for 3 pipelines at a time in a PR as you did for #272 after we merge #269 and #270.
Let me know on each PR when you're ready to merge them.

neodark:
The csv here still needs to be updated with the python script I guess

gperonato:
Ok, perfect, thanks.
This has been rebased and is ready to be merged.
The CSV has not been updated, as the python script was already included somehow in the master version.

neodark (approved):

neodark:
Ok I'll merge this one then for now.

PR #271, opened 2021-08-09
Integration pipeline for  get electricity

gperonato:


neodark:
Same comment as https://github.com/enermaps/enermaps/pull/269#issuecomment-895923437
Please rebase on master this PR.
Take into account that the db stuff has been split in a separate docker-compose for more clarity. The README in data-integration with this PR needs to be updated accordingly and re-tested if it still works. (There's now a `-f ../docker-compose-db.yml` before the python script to run)

gperonato:
Yes, good point. They were not modified here, but the master should take into account the changes from all pull requests.
Should I merge here the changes in the other PR so that there are no conflicts afterwards?

neodark (approved):

PR #276, opened 2021-08-13
Add a way to create a dump of the database within the docker container

gperonato:


neodark:
Note sure I understand this PR.
The pipelines help to fill the docker db container (local solution but not really usable as all pipelines could take multiple weeks to run).
So why a dump of this local database from the docker container ? (which may be incomplete)
As this local db container is not really filled, wouldn't this make more sense if the dump targets the real database in use with the enermaps server? (which should contain the real database)?

I understand the purpose of wanting to generate a dump from local data though.
Maybe we could keep this and add extra information on how to proceed with a dump from the `production` database.

gperonato:
Thank you for the comment! The purpose is actually to create the dump from the docker in the production database.
But to do so, the idea was to run  `pg_dump`within the docker container and save the file in a mounted volume, so that the  `pg_dump` version corresponds to the one of the database  (which is not necessarily the same as the one of the host machine).

neodark:
Ok then this was not clear. Maybe you could add more information that is to be done on the server side (i.e. ssh username@serverurl, then do this action, etc. etc.)

Or another proposition could be to indicate how to proceed with a `remote` `pg_dump` via ssh to a local mounted volume.

gperonato:
Thank you! I have integrated the README.

neodark (approved):

PR #279, opened 2021-08-16
Adding Nginx and Postgrest services to serve the gateway and the frontend

gperonato:


gperonato:
The continuous integration does not succeed because of the missing `.postgrest` file that is not synced with git.
Let me know if we can proceed anyway.

docker-compose-db.yml
```diff
@@ -26,6 +26,24 @@ services:
     env_file:
       .env-db
 
+  nginx:
```
neodark: db_nginx would be a better name otherwise one would think that it's the production server of the website

docker-compose-db.yml
```diff
@@ -26,6 +26,24 @@ services:
...
+    build:
+      context: ./nginx
```
neodark: It would be better to change the context as well to db_nginx (folder related to it) to not mix up with the website for example

nginx/README.md
```diff
@@ -0,0 +1,11 @@
+# nginx
```
neodark: same comment as above

gperonato:
> The continuous integration does not succeed because of the missing `.postgrest` file that is not synced with git.
> Let me know if we can proceed anyway.

The file is not called in the Dockerfile, so this does not seem a problem

docker-compose-db.yml
gperonato: Is this necessary? I think that the nginx server could be used for both (DB and frontend).
The root of the webserver (https://lab.idiap.ch/enermaps) could redirect for example to the frontend (`localhost:7000`), if we keep together all the services in the same `docker-compose.yml` as it was at the beginning

neodark: Actually yes it is necessary.
For multiple reasons:
1) the services could run on different servers: for example the db is currently on nginx while the website could use Apache
2) if the website is handled by someone else after the deployment, this could be a security issue if no access restrictions are present for the different services
3) it's a requirement to separate both services
4) it would become difficult for the IT team to calculate the effective cost and resources usage for each service if no proper separation is in place (specially taking into account that the website is managed by a different entity which is the CREM here)
5) at the moment for the deployment of the website no decision is taken with respect to what WSGI and server will be used, what are the security requirements, etc.

db/init/postgrest.sql
```diff
@@ -112,11 +112,12 @@ CREATE OR REPLACE FUNCTION enermaps_query_geojson(parameters text,
                     jsonb_object_agg(variable, value) as variables,
                     jsonb_object_agg(variable, unit) as units,
                     fields,
+                    layer,
```
melhk, line 115: What is this new variable?

gperonato: This is a json field containing information on the type of the layer so as to build a color bar/legend for categorical datasets. By default `{"type": "numerical"}` otherwise with classes, e.g.   {`"type": "categorical", "classes": {0: "warmer climate", 1: "average climate", 2: "colder climate"}}`

gperonato:
Hi @melhk @neodark, do you have any other comments/suggestions?

melhk:
> 
> 
> Hi @melhk @neodark, do you have any other comments/suggestions?

Flavio is reviewing this pull request but is not available these days

postgrest/Dockerfile
```diff
@@ -1 +1 @@
-FROM postgrest/postgrest
+FROM postgrest/postgrest:v7.0.1
```
melhk, line 1: Why don't you use the latest version (8.0.0)?

gperonato: It is not working for some reason. All the development tests were done on this version (which was the latest version till some weeks ago). I guess we can bump it to v8.0.0 later on with another PR

neodark: Ok @gperonato will let you check see that in another PR.

neodark (approved):

PR #292, opened 2021-08-27
Utilities to run cron jobs

gperonato:
This is used to run periodical jobs using the docker containers, notably the data-integrationpipelines that are run biannually to update the database content.
Please note that some of the pipelines run by the bash file are not yet included in the `master` branch.


neodark:
For your information, you could also use systemd which is more robust compared to cronjob as crontab doesn't allow tracking possible issues in the executed script.

neodark (approved):

PR #297, opened 2021-08-30
Case insensitive layer filtering

melhk:


neodark (approved):
Good catch! Thanks for the patch :+1:

PR #331, opened 2021-09-13
Fix metadata raster band index

gperonato:


neodark:
Will merge your fix, but just as a side note for general remarks:

1) Using a single letter as a variable can be confusing even if here we are just iterating.
2) Why the incrementation of that variable is done at the top of the loop?
3) Can we have edge cases where `b-1` wouldn't work ?

neodark (approved):

PR #340, opened 2021-09-21
Fix linting tools versions

Kanma:
Also reformat the SVELTE files

### DIFF
--- a/frontend/src/src/components/Map.svelte
+++ b/frontend/src/src/components/Map.svelte
@@ -1,140 +1,147 @@
 <script>
-import {onMount} from 'svelte';
-// Import CSS from Leaflet and plugins.
-import 'leaflet/dist/leaflet.css';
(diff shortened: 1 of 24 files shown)


frontend/src/src/components/Map.svelte
```diff
@@ -1,140 +1,147 @@
...
+  });
+
+  function resizeMap() {
+    if (map) {
+      map.invalidateSize();
+    }
+  }
+
+  $: {
+    console.log(`selected layer was changed: ${activeSelectionLayer}`);
```
neodark, line 57: On another PR, I would suggest to remove all those `console.log()` everywhere from the code.

Kanma: Either that or fix them, since most print something like: `[object Object]`

PuckCh (approved):

PR #342, opened 2021-09-24
Cm outputs cleanup

Kanma:
This PR ensures that the output files written by the CMs are deleted if not accessed for some time (set at 1 hour). Should an user need one file after its deletion, the frontend will start the CM again, using the original parameters.

As a side-effect, the logging done in the console of the web browser was improved to produce more useful debugging informations and remove the irrelevant ones.

### DIFF
--- a/cm/cm_heat_demand/tools/geofile.py
+++ b/cm/cm_heat_demand/tools/geofile.py
@@ -1,6 +1,6 @@
 import json
 import tempfile
-from os import close, remove, system
+from os import remove, system
 from os.path import exists, isfile, splitext
 
 import numpy as np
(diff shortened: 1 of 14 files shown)


cm/cm_heat_demand/tools/geofile.py
```diff
@@ -37,14 +38,18 @@ def clip_raster(src: str, shapes: dict, dst: str, quiet: bool = True):
         * projection : projection of the clipped raster.
     """
 
-    cutline = "cutline.json"
+    (file, cutline) = tempfile.mkstemp()
+    close(file)
+
     with open(cutline, "w") as file:
         json.dump(shapes, file)
```
PuckCh, line 43: Maybe something like:
```python
handle, cutline = tempfile.mkstemp()
with open(handle, "w") as f:
    json.dump(shapes, f)
```
is more simple to understand and don't open/close the file two times

Kanma: Right.

Today I learned something about open()

PuckCh (approved):

PR #344, opened 2021-09-27
Docker volumes

Kanma:
This PR does two things:

1) Use named volumes in the ```docker-compose``` configuration. Previously, new volumes with random names where created at each startup, for the ```frontend```, ```redis``` and ```flower``` containers.

2) Use separate volumes for the datasets (named ```geodb```) and the results of the CMs (```cm_outputs```), previously all stored in a single ```upload-data``` one. Additionally, the ```geodb``` volume is mounted as read-only on the CM containers.


Kanma:
Note that due to this change, you'll have to download the datasets again.

PuckCh (approved):

PR #255, opened 2021-07-01
CM HeatLearn

gperonato:
Hi @fallahnejad,
Could you do a general review of the CM architecture and features?
@melhk will later do a more thorough code review before the integration in the master branch.

Thank you all!

PR #354, opened 2021-10-06
CM HeatLearn

gperonato:
This is a new version of the CM adapted to the latest changes in the frontend and API.
- [x] Address comments from previous PR https://github.com/enermaps/enermaps/pull/255
- [x] Use the new endpoint to post rasters
- [x] Reading env variables
- [x] Provide a legend to the frontend
- [x] Accept more rasters (to be merged in one)


cm/heatlearn/heatlearn.py
```diff
@@ -23,13 +23,16 @@
 logging.basicConfig(level=logging.INFO)
 
 # REST API for HDD
-if not os.path.exists("API_KEY.txt"):
-    raise FileNotFoundError("Missing API key.")
-with open("API_KEY.txt", "r") as f:
-    API_KEY = f.read().replace("\n", "")
+env_db_server = ".env-db-server"
```
gperonato, line 26: Reading this file manually, as the env variables are not visible during testing.

gperonato:
The latest commit is a follow up of the discussion on Mattermost with @PuckCh. Cf. https://github.com/enermaps/enermaps/pull/354#discussion_r724048431

gperonato:
I think this is ready for review.
We could also merge and add the support to legends later.

cm/heatlearn/heatlearn.py
```diff
@@ -0,0 +1,345 @@
...
+# REST API for HDD
+env_db_server = ".env-db-server"
+if not os.path.exists(env_db_server):
+    raise FileNotFoundError("Missing .env-db-server file")
+with open(env_db_server, "r") as f:
+    for line in f.read().splitlines():
+        if line.startswith("DATASETS_SERVER_API_KEY"):
+            API_KEY = line.split("=")[-1].replace('"', "")
+
+POSTGREST_URL = "https://lab.idiap.ch/enermaps/api/db/rpc/enermaps_query_table"
```
Kanma: You should use the `DATASETS_SERVER_URL` environment variable from the file `.env-db-server` to build the `POSTGREST_URL` URL

cm/heatlearn/heatlearn.py
```diff
@@ -0,0 +1,345 @@
...
+API_URL = os.environ.get("API_BASE_URL")
```
Kanma: The `API_URL` variable doesn't seem to be used anywhere, so why the addition of a `API_BASE_URL` into `.env`?

cm/heatlearn/Dockerfile
```diff
@@ -0,0 +1,25 @@
+FROM ubuntu:20.04
+# SET TZ for GDAL
+ENV TZ=Europe/Zurich
+RUN ln -snf /usr/share/zoneinfo/$TZ /etc/localtime && echo $TZ > /etc/timezone
+
+RUN apt-get update && \
```
Kanma: You should do a `chmod 1777 /tmp && \` before `apt-get` here, like for the other CMs: it seems that the `ubuntu:20.04` image doesn't always a writable `/tmp` folder (it causes problem on the production server)

cm/heatlearn/heatlearn.py
gperonato: Indeed, it is only used in the `worker.py` and passed to the heatlearn() function, so I should remove it from here.

gperonato:
The CM is now working with the legends.

Kanma (approved):

PR #363, opened 2021-10-12
Fix legends 

gperonato:
Fix colors and labels in
- Default layers
- ESM pipeline


gperonato:
We should probably restart the production service when merged.

Kanma (approved):

PR #366, opened 2021-10-13
Add a first version of the CM on HDDs and CDDs

zarch:
A new clean PR to include the CM on HDDs and CDDs scenarios on the master.


zarch:
@cedricserugendo once again the new PR is failing in the lint check. However when I did the commit the pre-commit checks were passing:

```bash
$ git commit -m "Add a first version of the CM on HDDs and CDDs"
isort....................................................................Passed
black....................................................................Passed
Check python ast.........................................................Passed
Check for case conflicts.................................................Passed
Trim Trailing Whitespace.................................................Passed
Fix End of Files.........................................................Passed
Debug Statements (Python)................................................Passed
Check for added large files..............................................Passed
Check docstring is first.................................................Passed
Check Yaml...............................................................Passed
flake8...................................................................Passed
bandit...................................................................Passed
eslint...............................................(no files to check)Skipped
- hook id: eslint
[cm/hdd_cdd c6a83b9] Add a first version of the CM on HDDs and CDDs
 33 files changed, 642 insertions(+)
 create mode 100644 cm/cm_hdd_cdd/Dockerfile
 create mode 100644 cm/cm_hdd_cdd/README.md
 create mode 100644 cm/cm_hdd_cdd/hddcdd.py
 create mode 100644 cm/cm_hdd_cdd/requirements.txt
 create mode 100644 cm/cm_hdd_cdd/schema.json
 create mode 100644 cm/cm_hdd_cdd/setup.cfg
 create mode 100644 cm/cm_hdd_cdd/test.py
 create mode 100644 cm/cm_hdd_cdd/testdata/historical/cdd/22.0/monthly/average/2021_01.tif
 [...]
 create mode 100644 cm/cm_hdd_cdd/testdata/historical/hdd/18.0/monthly/average/2021_12.tif
 create mode 100644 cm/cm_hdd_cdd/worker.py
```

From the log of the PR precheck it is not clear what it is failing and what should I do to fix it.
Any hints?

cm/cm_hdd_cdd/test.py
```diff
@@ -0,0 +1,130 @@
...
+            t_base_h=18.0,
+            t_base_c=22.0,
+        )
+        from pprint import pprint
+
+        pprint(res)
+
+
+if __name__ == "__main__":
+    orig_repo = os.environ.get("CM_HDD_CDD_REPOSITORY", None)
```
cedricserugendo, line 127: this line seems not to be mandatory

cm/cm_hdd_cdd/worker.py
```diff
@@ -0,0 +1,42 @@
...
+    # print("=" * 50)
+    # print("CM-HDD-CDD")
+    # print("Selection:")
+    # pprint(selection)
+    # print("Rasters:")
+    # pprint(rasters)
+    # print("Params:")
+    # pprint(params)
+
+    # self.validate_params(params)
```
cedricserugendo: this line should not be commented I think

cm/cm_hdd_cdd/hddcdd.py
```diff
@@ -0,0 +1,214 @@
...
+        method="average",
+    )
+    avg_hdds = extract_by_dir(gdir=hdd_path, lon=lon, lat=lat)
+    cdd_path = get_datadir(
+        datarepository=get_datarepodir(),
+        sim_type=rcp,
+        dd_type="cdd",
+        Tb=t_base_c,
+        aggr_window="monthly",
+        method="average",
```
cedricserugendo: these values are default values, maybe for the readability of the code we should not specify them here again

cm/cm_hdd_cdd/hddcdd.py
```diff
@@ -0,0 +1,214 @@
...
+    print(f"CENTROID: {compute_centroid(geojson)}")
+    lon, lat = compute_centroid(geojson)
+    # Query the file-directory-netcdf structure for all the pixels involved
+    hdd_path = get_datadir(
+        datarepository=get_datarepodir(),
+        sim_type=rcp,
+        dd_type="hdd",
+        Tb=t_base_h,
+        aggr_window="monthly",
+        method="average",
```
cedricserugendo: these values are default values, maybe for the readability of the code we should not specify them here again

cm/cm_hdd_cdd/hddcdd.py
```diff
@@ -0,0 +1,214 @@
...
+    # select the right layer and return the results
+    # start = time()
+    msg = (
+        f"    » ref. year: {refyear},"
+        f" crp: {rcp}, Tbh: {t_base_h:.1f}, Tbc: {t_base_c:.1f}"
+    )
+    logging.info(msg)
+    # Compute the centroid of the geometry selected by the user
+    print(f"CENTROID: {compute_centroid(geojson)}")
+    lon, lat = compute_centroid(geojson)
```
cedricserugendo: maybe you should save the value returned by the `compute_centroid`, so you don't perform the fuction twice

cm/cm_hdd_cdd/hddcdd.py
```diff
@@ -0,0 +1,214 @@
...
+    return sr
+
+
+def dd_my_group(sr: pd.Series) -> pd.Series:
+    isplit = sr.index.str.split("_")
+    years = [yr for yr, _ in isplit]
+    mnths = [mnth for _, mnth in isplit]
+    yrgrp = sr.groupby(years)
+    mngrp = sr.groupby(mnths)
+    return mngrp, yrgrp
```
cedricserugendo: you don't use this function I think

cm/cm_hdd_cdd/hddcdd.py
```diff
@@ -0,0 +1,214 @@
...
+    res = []
+    idx = []
+    for gfi in gdir.iterdir():
+        if gfi.name.endswith(".tif"):
+            idx.append(gfi.name[:-4])
+            try:
+                gx = __datasets[gfi.as_posix()]
+            except KeyError:
+                gx = rasterio.open(gfi)
+                __datasets[gfi.as_posix()] = gx
```
cedricserugendo, line 92: I suggest this solution to close the file properly 
```python
try:
     gx = __datasets[gfi.as_posix()]
except KeyError:
     with rasterio.open(gfi) as file:
           gx = file
           __datasets[gfi.as_posix()] = gx
```

cm/cm_hdd_cdd/hddcdd.py
```diff
@@ -0,0 +1,214 @@
...
+
+
+def scale_stat(stats: dict, factor):
+    """From a list of stats, take the factor into account and
+    modify the stats accordingly.
+    """
+    for stat_name, stat in stats.items():
+        for scaled_stat_prefix in SCALED_STATS_PREFIX:
+            if stat and stat_name.startswith(scaled_stat_prefix):
+                stats[stat_name] = stat * factor
```
cedricserugendo: I think that you don't use this function

cm/cm_hdd_cdd/hddcdd.py
```diff
@@ -0,0 +1,214 @@
...
+DEFAULT_STATS = ("min", "max", "mean", "median", "count")
+SCALED_STATS_PREFIX = ("min", "max", "mean", "median", "percentile_")
+CURRENT_FILE_DIR = os.path.dirname(os.path.abspath(__file__))
+
+CDF_POINTS = range(0, 101)
+
+DECIMALS = 3
+
+CURRENT_FILE_DIR = Path(__file__).parent
+TESTDATA_DIR = CURRENT_FILE_DIR / "testdata"
```
cedricserugendo, line 15: you redeclare the variable `CURRENT_FILE_DIR`, may be you can keep the declearation of the line 15

cm/cm_hdd_cdd/hddcdd.py
```diff
@@ -0,0 +1,214 @@
...
+import pandas as pd
+import rasterio
+from BaseCM.cm_output import validate
+
+GEOJSON_PROJ = "EPSG:4326"
+DEFAULT_STATS = ("min", "max", "mean", "median", "count")
+SCALED_STATS_PREFIX = ("min", "max", "mean", "median", "percentile_")
+CURRENT_FILE_DIR = os.path.dirname(os.path.abspath(__file__))
+
+CDF_POINTS = range(0, 101)
```
cedricserugendo: you don't use this variable I think

cm/cm_hdd_cdd/hddcdd.py
```diff
@@ -0,0 +1,214 @@
...
+from typing import Dict, Tuple
+
+import numpy as np
+import pandas as pd
+import rasterio
+from BaseCM.cm_output import validate
+
+GEOJSON_PROJ = "EPSG:4326"
+DEFAULT_STATS = ("min", "max", "mean", "median", "count")
+SCALED_STATS_PREFIX = ("min", "max", "mean", "median", "percentile_")
```
cedricserugendo: you don't use this variables I think

cedricserugendo (changes requested):
Just some small suggestions 
I can't run the code locally so most of my comments are about form 
The error code I get is the following 
```bash
cm-hdd_cdd_1      |   File "/usr/local/lib/python3.8/dist-packages/celery/app/trace.py", line 409, in trace_task
cm-hdd_cdd_1      |     R = retval = fun(*args, **kwargs)
cm-hdd_cdd_1      |   File "worker.py", line 30, in cm_hddcdd
cm-hdd_cdd_1      |     res = hdd_cdd_stats(
cm-hdd_cdd_1      |   File "/cm-hdd_cdd/hddcdd.py", line 176, in hdd_cdd_stats
cm-hdd_cdd_1      |     avg_hdds = extract_by_dir(gdir=hdd_path, lon=lon, lat=lat)
cm-hdd_cdd_1      |   File "/cm-hdd_cdd/hddcdd.py", line 102, in extract_by_dir
cm-hdd_cdd_1      |     for gfi in gdir.iterdir():
cm-hdd_cdd_1      |   File "/usr/lib/python3.8/pathlib.py", line 1122, in iterdir
cm-hdd_cdd_1      |     for name in self._accessor.listdir(self):
cm-hdd_cdd_1      | FileNotFoundError: [Errno 2] No such file or directory: '/cm-hdd_cdd/testdata/4.5/hdd/18.0/monthly/average'
```
The error probably comes from how the path is created 
But I don't find a solution, do you have any idea how I can get rid of this error?

And the structure of the code to has changed, we should rebase this branch from master

I am at your disposal if you have any questions @zarch

cm/cm_hdd_cdd/hddcdd.py
zarch: You are right, the following line will be removed:

```python
-GEOJSON_PROJ = "EPSG:4326"
-DEFAULT_STATS = ("min", "max", "mean", "median", "count")
-SCALED_STATS_PREFIX = ("min", "max", "mean", "median", "percentile_")
```

cm/cm_hdd_cdd/hddcdd.py
zarch: Right, I removed the `CDF_POINTS`

cm/cm_hdd_cdd/hddcdd.py
zarch: Correct, I removed the double declaration.

cm/cm_hdd_cdd/hddcdd.py
zarch: Correct, removed.

cm/cm_hdd_cdd/hddcdd.py
zarch: I cannot close the file, the local variable: `__datasets: Dict[str, rasterio.DatasetReader] = {}` is creating a cache with `rasterio.DatasetReader` instances, to avoid to open the file every time the server is receiving a request from a client. If I close the rasterio instance than I would not be able to query the raster later using the `gx` variable:

```python
@lru_cache()
def extract_by_dir(
    gdir: Path,
    lon: float,
    lat: float,
    __datasets: Dict[str, rasterio.DatasetReader] = {},
):
    res = []
    idx = []
    for gfi in gdir.iterdir():
        if gfi.name.endswith(".tif"):
            idx.append(gfi.name[:-4])
            try:
                gx = __datasets[gfi.as_posix()]
            except KeyError:
                gx = rasterio.open(gfi)
                __datasets[gfi.as_posix()] = gx
            yp, xp = gx.index(lon, lat)      # <=here I'm using the raterio gx instance
            val = gx.read()[0][yp, xp]       # <=here I'm using the raterio gx instance
            logging.info(
                f"{gfi.name}@{lon:.{DECIMALS}f},{lat:.{DECIMALS}f}: {yp}, {xp} => {val}"
            )
            res.append(val)
    sr = pd.Series(np.array(res), index=idx, name=f"yp={yp},xp={xp}")
    sr.sort_index(inplace=True)
    return sr
```
Open the file without closing it shouldn't be a big issue here, because:
* we are opening the file in read-only mode;
* we are doing so only once per server process not per client request

We might need to raise the docker open file limits,  the default value in ubuntu is: 176772.

cm/cm_hdd_cdd/hddcdd.py
zarch: You are correct, I removed it.

cm/cm_hdd_cdd/hddcdd.py
zarch: The issue here is that I receive a geojson as a Dict, therefore I cannot create a cache because dict in an unhashable type. So I should convert the geojson dict back to string and then I can create a cache, but I think that doing so I'm going to loose most of the time that I might save with the cache. To compute the centroid, I'm just:
* creating a numpy array instance,
* compute the mean values, 
* round the results and 
* transform the output into a tuple.
I think this function should not be that critical. However if you think it is, then we should create a good benchmark function that use real geojson dictionaries from our NUTS{0,1,2,3} and LAU2 to see which is the real gain in adding a cache at this point.

zarch: aah, now I see what you mean, here I'm calling the function two times, ok, I've removed the call in the print statement, now is called just once.

cm/cm_hdd_cdd/hddcdd.py
zarch: ok, removed

cm/cm_hdd_cdd/hddcdd.py
zarch: ok, done

cm/cm_hdd_cdd/test.py
zarch: I'm not sure that I correctly understood which line do you mean, are you referring to this one:
`orig_repo = os.environ.get("CM_HDD_CDD_REPOSITORY", None)`
Here, I'm saving the original value, because I'm changing the user defined value to do the tests, once the test are done I'm setting back the original value.

cm/cm_hdd_cdd/worker.py
zarch: I comment them out otherwise the log is quite verbose for each client request, but I keep it as comment if someone need to locally test what the function is receiving from the server.

zarch:
> Just some small suggestions I can't run the code locally so most of my comments are about form The error code I get is the following
> 
> ```shell
> Traceback (most recent call last):
> cm-hdd_cdd_1      |   File "/usr/local/lib/python3.8/dist-packages/celery/app/trace.py", line 409, in trace_task
> cm-hdd_cdd_1      |     R = retval = fun(*args, **kwargs)
> cm-hdd_cdd_1      |   File "worker.py", line 30, in cm_hddcdd
> cm-hdd_cdd_1      |     res = hdd_cdd_stats(
> cm-hdd_cdd_1      |   File "/cm-hdd_cdd/hddcdd.py", line 176, in hdd_cdd_stats
> cm-hdd_cdd_1      |     avg_hdds = extract_by_dir(gdir=hdd_path, lon=lon, lat=lat)
> cm-hdd_cdd_1      |   File "/cm-hdd_cdd/hddcdd.py", line 102, in extract_by_dir
> cm-hdd_cdd_1      |     for gfi in gdir.iterdir():
> cm-hdd_cdd_1      |   File "/usr/lib/python3.8/pathlib.py", line 1122, in iterdir
> cm-hdd_cdd_1      |     for name in self._accessor.listdir(self):
> cm-hdd_cdd_1      | FileNotFoundError: [Errno 2] No such file or directory: '/cm-hdd_cdd/testdata/4.5/hdd/18.0/monthly/average'
> ```
> 
> The error probably comes from how the path is created But I don't find a solution, do you have any idea how I can get rid of this error?

Do you mean if you execute locally on your computer and you are selecting an option that is different than `historical`, `18.0` and `22.0`?

Yes it is failing because at the moment is using the `cm/cm_hdd_cdd/testdata` directory as default repository. 
I'm still finalizing the dataset creation, once is finalized I will download the dataset  when building the Docker, and I will set the repository path to the `CM_HDD_CDD_REPOSITORY` environmental variable.

zarch:
> And the structure of the code to has changed, we should rebase this branch from master

I'm going to rebase it now.

cm/cm_hdd_cdd/test.py
cedricserugendo: ok I didn't understand that you were changing this value, in this case everything is ok

cm/cm_hdd_cdd/worker.py
cedricserugendo: ok, let's keep these lines as comments but I think it would be a good idea to uncomment this line `self.validate_params(params)`

cedricserugendo (changes requested):

cm/cm_hdd_cdd/input_layers.json
```diff
@@ -0,0 +1 @@
+[]
```
zarch: @Kanma I see that you add the documentation in the README file about the input_layers.json.
In my case the CM do not need any file to be selected from the user.
I think that use: 
```json
[
    {
        "dataset": "all",
    }
]
```
It is wrong because it require the user to select a layer that then is not used by the CM, therefore I think there should be a way to declare that the CM do not need any layer and user have to select only the region of interest.
What do you think? Should I open an issue?

Kanma: Hello, I'll make the necessary changes to support that.

Kanma: @zarch, the following changes are now in the `master`branch:

(from the README)

```
`dataset` is the ID of the dataset. The following special values are available:

* "all" can be used to specify all the datasets
* "all_rasters" can be used to specify all the raster datasets
* "all_vectors" can be used to specify all the vector datasets
* "none" can be used to indicates that the CM doesn't require any dataset at all
```

PR #369, opened 2021-10-14
add string processing

cedricserugendo:


cedricserugendo:
I added an argument to black so that, during the precommit, it processes the strings.

Kanma (approved):

PR #380, opened 2021-10-26
Update postgrest.sql

gperonato:
Small fix


Kanma:
This one is already in `master`, so I'm closing it.

PR #394, opened 2021-10-27
Only send to CMs the raster files that intersect the selection area

Kanma:
Same strategy than for the map creation in the WMS


Kanma:
Don't accept it now, I found a bug

Kanma:
Fixed

PuckCh (approved):

PR #406, opened 2021-11-05
Integrate wiki platform source

peter-gy:
### Goals

- Integrate the wiki platform developed under [enermaps/doc/dev](https://github.com/enermaps/doc/tree/dev), as discussed with @gperonato
- Develop the platform with automated translation compatibility in mind
- Authenticate wiki editors via GitHub on the platform
  - For this purpose I set up a new OAuth App for the organisation, accessible under the developer settings

### Notes

- The [enermaps/doc/data](https://github.com/enermaps/doc/tree/data) branch serves currently as the remote "database" of the wiki. It is necessary to handle content synchronisation when running translations automatically

PR #413, opened 2021-11-08
Enhance the documentation platform (ubuntu:20.04 & gollum:5.2.3)

peter-gy:
### Goals

- Follow up of #406
- Fix the problem with replacing existing images by bumping the version of gollum
- Finalize documentation

### Notes

- The platform is already live under https://enermaps-wiki.herokuapp.com/ - it was deployed directly from this source
- The documentation platform runs currently on the most recent version of gollum, coming with various enhancements
- While reworking the `Dockerfile`, I considered the comment of @gperonato and composed the base container from `ubuntu:20.04` to stay consistent stack-wise with the rest of the project
- Existing uploads can be managed under https://enermaps-wiki.herokuapp.com/gollum/overview/uploads/


gperonato (approved):
Thank you for this update!

PR #411, opened 2021-11-08
set zoom levels

gperonato:

PR #414, opened 2021-11-09
Use minimum zoom levels defined in the database instead of a fixed one

Kanma:
Those info were added by PR #411


gperonato:
Tested with #411 values.

PuckCh (approved):

PR #421, opened 2021-11-12
raster statistics 

cedricserugendo:


Kanma:
This is a replacement for `example_multiply` right?

PR #423, opened 2021-11-15
CM input layers

Kanma:
Add new options to specify the layers a CM can work on: "all_rasters", "all_vectors" and "none"


gperonato:
Thank you. I think the `none`option is what is needed by @zarch.

gperonato (approved):
For being future-proof, it would be useful to be able to select also the result of a CM as the input of another CM. I was thinking that the `heat demand` CM by @cedricserugendo  could be theoretically run on the output of the heatlearn cm.

PR #447, opened 2022-02-01
add a common module to all CMs for the management of rasters

cedricserugendo:


cedricserugendo:
I have added a function that could be common to all CMs in the same place

Kanma:
Shouldn't the functions be removed from `cm/cm_heat_demand/tools/geofile.py`?

cedricserugendo:
> Shouldn't the functions be removed from `cm/cm_heat_demand/tools/geofile.py`?

yes indeed, I thank you for your remark @Kanma 
I made the changes

Kanma (approved):

PR #448, opened 2022-02-01
add multiple chart option

cedricserugendo:


cedricserugendo:
I added the possibility to return several graphs in CM output

Kanma (approved):

PR #451, opened 2022-02-17
CM: hdd_cdd main refactoring

zarch:
The main changes on this PR are dealing with these main folders:
- `cm/base` where:
    - modify the `setup.py` file and 
    - add a `MANIFEST.in` file to also include the GeoTiff data used for testing;
    - change the `cm/base/requirements.txt`to include pandas and geopandas that are required by `cm/base/BaseCM/cm_hddcdd.py`;
-  `cm/base/BaseCM` where:
    - I moved the shared functions between the `hdd-cdd`and `refurbish` CM on a dedicated file: `cm/base/BaseCM/cm_hddcdd.py`. I've call it in that way for consistency with the other files;
    - I add the new version of the GeoTiff that is an extraction of the full dataset available in the [repository](https://gitlab.inf.unibz.it/URS/enermaps/hdd-cdd);
    - I add more testing
- `cm/hdd_cdd` more or less use the standard structure, with two main exceptions:
   -  the `schema.json` file that is generated automatically during the building process based on the data that has been downloaded;
   -  the `docker-compose`definition of the service that create a volume in `rw`mode to avoid to duplicate the shared data between the `hdd_cdd`and the `refurbish` CM.


cm/base/BaseCM/cm_hddcdd.py
```diff
@@ -0,0 +1,271 @@
...
+    # yrs = get_years()
+    # refyr = dict(
+    #     type="integer",
+    #     title="Reference year",
+    #     description="",
+    #     default=2020,
+    #     minimum=min(yrs),
+    #     maximum=max(yrs),
+    #     enum=yrs,
+    # )
```
cedricserugendo: Can we delete this part?

cm/base/BaseCM/cm_hddcdd.py
```diff
@@ -0,0 +1,271 @@
...
+            json.dump(schema, schfile, indent=2, sort_keys=True)
+    return schema
+
+
+def download_data():
+    # export CM_HDD_CDD_REPOSITORY=/tmp/hddcddrepo
+    # export INPUT_DATA_DIR=/tmp
+    #  /cm_inputs/hdd-cdd/
+    rdir = get_data_dir()
+    #  /cm_inputs/hdd-cdd/hdd-cdd-main/data
```
cedricserugendo: Can we delete the commented line ?

cm/base/BaseCM/test_hddcdd.py
```diff
@@ -0,0 +1,201 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
```
cedricserugendo, line 2: These tests are related to the CM hddcdd, ideally they should be in the directory of this CM 
If the tested features are used by several CMs, they will just have to be tested several times (maybe with different data depending on how the CM works)

cm/base/BaseCM/test_hddcdd.py
```diff
@@ -0,0 +1,201 @@
...
+        gj = deepcopy(GEOJSON)
+        geo = gpd.GeoDataFrame.from_features(gj["features"], crs="EPSG:4326").geometry
+        cx, cy = hc.compute_centroid(geo)
+        self.assertAlmostEqual(cx, round(11.04713688, ndigits=hc.DECIMALS))
+        self.assertAlmostEqual(cy, round(45.56383191, ndigits=hc.DECIMALS))
+
+    def test__reproj(self):
+        # https://epsg.io/transform#s_srs=4326&t_srs=3035&x=11.0132789&y=45.5228261
+        # (lat, lon)
+        # (45.5228261, 11.0132789) => (4400277.98, 2490583.97)
```
cedricserugendo: These lines should be the docstring of `hc.reproj` function

cm/hdd_cdd/Dockerfile
```diff
@@ -0,0 +1,94 @@
...
+    && cd /tmp/base && python3 setup.py install \
+    && python3 test.py \
+    && echo "-----\n"
+
+
+## ---------------------------------
+## Download required data and layers
+## ---------------------------------
+## by default use the python function to download the missing data
+RUN python3 -c "from BaseCM import cm_hddcdd as hc; hc.download_data()"
```
cedricserugendo: it may be a separate script call by the Docker file (only for reading issue)

cm/hdd_cdd/Dockerfile
```diff
@@ -0,0 +1,94 @@
...
+## -----------------------------
+## Generate the schema.json file
+## -----------------------------
+# extract from the data set the main parameters and save them on the schema.json
+RUN python3 -c "from pprint import pprint; \
+    from pathlib import Path; \
+    from BaseCM import cm_hddcdd as hc; \
+    schema = hc.get_hddcdd_schema(save=True, schema_path=Path('schema.json')); \
+    print('Generated schema:'); \
+    pprint(schema)"
```
cedricserugendo: it may be a separate script call by the Docker file (only for reading issue)

cm/hdd_cdd/worker.py
```diff
@@ -0,0 +1,53 @@
...
+    If there are many rasters, we select the first one.
+    """
+    # logging.info("=" * 50)
+    # logging.info("CM-HDD-CDD")
+    # logging.info("Selection:")
+    # logging.info(selection)
+    # logging.info("Rasters:")
+    # logging.info(rasters)
+    # logging.info("Params:")
+    # logging.info(params)
```
cedricserugendo: Can we delete this part ?

cedricserugendo (commented):
I made some small remarks/suggestions
the linter problem is also solved, I'm at your disposal if needed

cm/base/BaseCM/test_hddcdd.py
zarch: No these test are related to all the functions that are in the `BaseCM.cm_hddcdd` module. To me it make sense to add a test here, so in the same folder of where I defined the function.
Please consider the the refurbish CM is also using these functions.

cedricserugendo: yes it makes sense, we can keep the codes as you wrote it

cm/hdd_cdd/worker.py
```diff
@@ -0,0 +1,45 @@
...
+    base=cm_base.CMBase,
+    bind=True,
+    schema_path=schema_path,
+    input_layers_path=input_layers_path,
+)
+def hdd_cdd(self, selection: dict, rasters: list, params: dict):
+    """This is a calculation module that multiplies the raster by an factor.
+    If there is no raster, we raise a value error.
+    If there are many rasters, we select the first one.
+    """
```
cedricserugendo: we should change this docstring

cedricserugendo (approved):

PR #453, opened 2022-02-24
CM: refurbish

zarch:
I changed the `BaseCM.cm_hddcdd` to:
* remove the `sys.exit()` because otherwise the download function was stopping with the first dataset;
* replace [`rasterio.read`](https://rasterio.readthedocs.io/en/latest/api/rasterio._io.html#rasterio._io.DatasetReaderBase.read)  with  [`rasterio.sample`](https://rasterio.readthedocs.io/en/latest/api/rasterio._io.html#rasterio._io.DatasetReaderBase.sample) to avoid to read and load into memory everytime the whole array to read just a single pixel value;
* replace the use of `dict`with the `OrderedDict` to respect the parameter order when generating the `schema.json` file;
* add `atexit.register` function to properly close the rasterio dataset before terminating the process.

Then I add the `cm/refurbish` folder respecting the standard structure. As for the `cm/hdd_cdd` also in the `cm/refurbish` we are generating the `scema.json` during the building process, through the `python schema.py` command.

Do we have already in the repository a function that convert a number changing the value to the best unit prefix? 
Something like: `1_000_000 Wh` => `1 MWh`. 

For Hotmaps I developed these functions [resutils/unit.py](https://github.com/HotMaps/resutils/blob/master/resutils/unit.py), what do you think it might be the best approach:
* include the `unit.py` file somewhere in the repo, most probably under `BaseCM` since I would expect that is a common functionality;
* put `reslib` within the `cm/refurbish/requirements`


cm/refurbish/refurbish.py
```diff
@@ -0,0 +1,643 @@
...
+    bs["end"] = ends
+    return bs
+
+
+def path_population() -> Path:
+    """Return the path to the population dataset."""
+    return cm_hddcdd.get_data_dir() / os.environ.get(
+        "POPGJSN", "LAU_RG_01M_2020_4326.geojson"
+    )
+
```
cedricserugendo: When I test your CM, I get the following error: 
```
/cm_inputs/hdd-cdd/LAU_RG_01M_2020_4326.geojson: No such file or directory
```
I think the error comes from the following function.
In the directory `/cm_inputs/hdd-cdd/`, there is only the directory `hdd-cdd-main`

cedricserugendo (changes requested):

cedricserugendo:
@zarch  
We don't have conversion functions for units.
I think the best solution is to import the function you wrote for HotMaps.

cedricserugendo (dismissed):

cedricserugendo (approved):

PR #460, opened 2022-03-10
add output layer type to schema json

cedricserugendo:


fallahnejad (changes requested):
I think you need to use the link that works on the EDMT and not the link for the Docker.

cedricserugendo:
@Kanma changed the `get_raster_path` function and the path I use should work for the local test as well as for the online version

fallahnejad (changes requested):
Please use correct link for the hdm_layer.

fallahnejad (approved):

PR #487, opened 2022-04-20
Legends

ManuelDalcaEurac:
The following resources have been modified/uploaded:
- .env file: inserted a variable API_KEY whose fake value should be substituted with the private key required to upload legends
- docker-compose-db.yml: inserted the "legends" service
- legends/: folder containing the required resources in order to run the script which updates enermaps legends

The load_legends.py script should read raster data directly from the server. I tested the code after running data-integration/getPVGIS.py (raster based) and data-integration/getJRC-PPDB-OPEN.py (vector based), and the legends for these two datasets are already visible on the platform. The scripts of the "data-integration" service save raster data in /data-integration/data/ds_id/fid, where /data-integration/data/ is shared with the "legends" service through docker, which is where load_legends.py reads the data from. The connection to the db through postgrest is the same as in other scripts. In the case of raster data, breaks are computed on a sample based on Quasi Random Sequences, which is more efficient wrt Pure Random Sampling. Breaks are saved on the docker volume (each file is less than 1kb), and then uploaded on the server as legends.


legends/load_legends.py
```diff
@@ -0,0 +1,309 @@
...
+                        "value": breaks[color_index],
+                        "label": "≥ "
+                        + str(round(breaks[color_index], 2))
+                        + " "
+                        + str(row["unit"]),
+                    }
+                )
+
+            # send the legend
+            r = requests.post(
```
gperonato, line 284: Since we connect anyway to the database service, I think it would be better to use the `enermaps_set_legend` directly as a SQL query. It should be faster and "cleaner". Maybe to be added to the README as something that can be considered by future users that implement their own EnerMaps.

legends/load_legends.py
```diff
@@ -0,0 +1,309 @@
...
+db_connection_url = (
+    "postgresql://{DB_USER}:{DB_PASSWORD}@{DB_HOST}:{DB_PORT}/{DB_DB}".format(
+        DB_HOST=DB_HOST,
+        DB_PORT=DB_PORT,
+        DB_USER=DB_USER,
+        DB_PASSWORD=DB_PASSWORD,
+        DB_DB=DB_DB,
+    )
+)
+
```
gperonato, line 90: Add a `if __name__ == "__main__"` here?

legends/load_legends.py
```diff
@@ -0,0 +1,309 @@
...
+# read colors and layer names
+colors_dict = get_colors(CLRTABLE)
+layers = pd.read_csv(LAYERTABLE)
+layers["uploaded"] = 0
+
+# ------------------------------
+# computing breaks
+# ------------------------------
+
+for i, row in layers.iterrows():
```
gperonato, line 111: It would be useful to be able to add an argument (e.g., the "dataset_id"), if we do not want to calculate all legends but only replace the ones for the layers that are updated.

gperonato (commented):
Thank your for your work!
Just some suggestions below.
I will test meanwhile the service in the server.

Can you also add a `README.md` file in the `legends` subdirectory and edit the README in the root directory?

legends/load_legends.py
gperonato: Something like this:
```
if len(sys.argv) > 1:
    ds_ids = [int(x) for x in sys.argv[1:]]
    layers = layers.loc[layers["ds_id"].isin(ds_ids),:]
```

.env
```diff
@@ -3,3 +3,6 @@ CELERY_BROKER_URL=redis://redis
 CELERY_RESULT_BACKEND=redis://redis
 API_URL=http://api/api
 API_BASE_URL=http://api
+
```
gperonato, line 6: I would move this to another env file `.env-legends`to be included in the `docker-compose.yml` file

legends/load_legends.py
```diff
@@ -0,0 +1,309 @@
...
+                        "label": "≥ "
+                        + str(round(breaks[color_index], 2))
+                        + " "
+                        + str(row["unit"]),
+                    }
+                )
+
+            # send the legend
+            r = requests.post(
+                "https://lab.idiap.ch/enermaps/api/db/rpc/enermaps_set_legend",
```
gperonato, line 285: This does not work if the service is run internally (e.g. if the API is also a docker service). I would suggest using this:
```            r = requests.post(
                "{}".format(POSTGREST_URL),
                headers={"Authorization": "Bearer {}".format(API_KEY)},
```
where the POSTGREST_URL variable is defined in the `.env-legends` file and retrieved at the beginning of the script with `os.environ.get()`

legends/load_legends.py
```diff
@@ -0,0 +1,309 @@
...
+            print(
+                'Computing legend breaks - dataset {}, layer "{}"\n'.format(
+                    ds_id, variable
+                )
+            )
+            vector = gdf
+            vector["variable"] = vector["variables"].apply(
+                lambda x: list(x.values())[0]
+            )
+            data = np.array(vector["variable"])
```
gperonato, line 223: For dataset 3, the `ValueError("The breaks are less than 2")` is raised.
I would suggest using this here:
`data = np.array(vector["variable"].dropna())`

gperonato (commented):
Some other comments while testing

